PR #21, opened 2023-10-31
BENCH-1233: Append workspace link in proxy agent banner

rogerwangcs:
Inject a workspace link in the dataproc component proxy banner. 

Screenshot 2023-10-31 at 5 07 19 PM

Ticket: https://verily.atlassian.net/browse/BENCH-1233

### DIFF
--- a/startupscript/dataproc/startup.sh
+++ b/startupscript/dataproc/startup.sh
@@ -986,18 +986,18 @@ fi
   sed -i 's#<banner-title>#<banner-title>\n'"${WORKSPACE_LINK_EL}"' \&gt; #' "${PROXY_AGENT_BANNER}"
 
   # Add target blank property to all banner links so they open in a new tab
-  sed -i "s/class=\"forum\"/class=\"forum\" target=\"_blank\"/g" ${PROXY_AGENT_BANNER}
+  sed -i 's#class="forum"#class="forum" target="_blank"#g' "${PROXY_AGENT_BANNER}"
 
   # Remove flex styling from the banner-account css class to prevent banner content from wrapping
-  sed -i '/banner-account {/,/}/{/flex:/d;/-ms-flex:/d;/-webkit-flex:/d;}' ${PROXY_AGENT_BANNER}
+  sed -i '#banner-account {#,#}#{#flex:#d;#-ms-flex:#d;#-webkit-flex:#d;}' "${PROXY_AGENT_BANNER}"
 
-  # Add css class for #workspace before a#project
+  # Add css class for a#workspace before a#project
   sed -i '/a#project {/i\
 a#workspace {\
   color:white;\
   text-decoration:none;\
   padding:4px;\
-}' ${PROXY_AGENT_BANNER}
+}' "${PROXY_AGENT_BANNER}"
 
   # End banner.html modifications
 
--- a/startupscript/dataproc/startup.sh
+++ b/startupscript/dataproc/startup.sh
@@ -966,7 +966,7 @@ fi
       ui_base_url=terra-preprod-ui-terra.api.verily.com
     ;;
     verily)
-      ui_base_url=workbench.verily.com/workspaces/
+      ui_base_url=workbench.verily.com
       ;;
     *)
       >&2 echo "ERROR: $TERRA_SERVER is not a known verily server."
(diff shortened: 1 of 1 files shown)


startupscript/dataproc/startup.sh
```diff
@@ -935,6 +947,58 @@ fi
...
+  # Add css class for #workspace before a#project
+  sed -i '/a#project {/i\
+a#workspace {\
+  color:white;\
+  text-decoration:none;\
+  padding:4px;\
+}' ${PROXY_AGENT_BANNER}
+
+  # restart proxy agent
+  systemctl restart ${PROXY_AGENT_SERVICE}
```
mbookman, line 996:
```
  systemctl restart "${PROXY_AGENT_SERVICE}"
```

startupscript/dataproc/startup.sh
```diff
@@ -935,6 +947,58 @@ fi
...
+    verily)
+      ui_base_url=workbench.verily.com
+      ;;
+    *)
+      >&2 echo "ERROR: $TERRA_SERVER is not a known verily server."
+      exit 1
+      ;;
+  esac
+
+  # Insert a link element to render
```
mbookman, line 977: Updating CSS embedded in HTML here is pretty fragile.
I know that there are often no _great_ choices for this kind of customization.

However, for changes like these one, we should at least consider writing a little bit of python that could take advantage of libraries like https://docs.python.org/3/library/html.parser.html.

Short of reworking this, please add a comment like

```
# The banner.html file contains <style> wrapper tags and a series of CSS styles,
# and a block of embedded javascript (inside a <script> tag) that we want to update.
```

startupscript/dataproc/startup.sh
```diff
@@ -935,6 +947,58 @@ fi
...
+  readonly WORKSPACE_LINK_EL="<a id=\"workspace\" class=\"forum\" target=\"_blank\" href=\"https://${ui_base_url}/workspaces/${TERRA_WORKSPACE}\">${TERRA_WORKSPACE}</a>"
```
mbookman, line 978: I think this reads easier as:

```
  readonly WORKSPACE_LINK_EL='<a id="workspace" class="forum" target="_blank" href="https://'"${ui_base_url}/workspaces/${TERRA_WORKSPACE}"'"'">${TERRA_WORKSPACE}</a>"
```

Though admittedly getting that last double-quote around the href is a little clunky.

startupscript/dataproc/startup.sh
```diff
@@ -935,6 +947,58 @@ fi
...
+  sed -i "s/<banner-title>/<banner-title>\n$WORKSPACE_LINK_EL \&gt; /" ${PROXY_AGENT_BANNER}
```
mbookman, line 979: I think it's preferable to use single quotes where possible, since we are often trying to emit double-quotes.
Also suggest using `#` instead of `/` since we're more likely to want to emit path symbols than hashes.

```
sed -i 's#<banner-title>#<banner-title>\n'"${WORKSPACE_LINK_EL}"' \&gt; #' "${PROXY_AGENT_BANNER}"
```

startupscript/dataproc/startup.sh
```diff
@@ -949,6 +961,53 @@ fi
...
+
+  # The banner.html file contains <style> wrapper tags and a series of CSS styles, and a set of html link elements that we want to modify.
+  # Begin banner.html modifications
+
+  # Insert a workspace link into the banner title
+  readonly WORKSPACE_LINK_EL='<a id="workspace" class="forum" target="_blank" href="https://'"${ui_base_url}/workspaces/${TERRA_WORKSPACE}"'"'">${TERRA_WORKSPACE}</a>"
+  sed -i 's#<banner-title>#<banner-title>\n'"${WORKSPACE_LINK_EL}"' \&gt; #' "${PROXY_AGENT_BANNER}"
+
+  # Add target blank property to all banner links so they open in a new tab
+  sed -i "s/class=\"forum\"/class=\"forum\" target=\"_blank\"/g" ${PROXY_AGENT_BANNER}
```
mbookman, line 989: I think this reads better as:

```
  sed -i 's#class="forum"#class="forum" target="_blank"/g' "${PROXY_AGENT_BANNER}"
```

startupscript/dataproc/startup.sh
```diff
@@ -949,6 +961,53 @@ fi
...
+
+  # Remove flex styling from the banner-account css class to prevent banner content from wrapping
+  sed -i '/banner-account {/,/}/{/flex:/d;/-ms-flex:/d;/-webkit-flex:/d;}' ${PROXY_AGENT_BANNER}
```
mbookman, line 992: I think it'd be preferred to use hashes to separate the substitution matching, since there's a lot of slashes in there. At the very least, the `${PROXY_AGENT_BANNER}` should be quoted.

mbookman (changes requested):

startupscript/dataproc/startup.sh
rogerwangcs: Done.

mbookman (approved):
LGTM

PR #25, opened 2023-11-06
add jupyter config later in the script

yuhuyoyo:


startupscript/dataproc/startup.sh
```diff
@@ -752,9 +746,6 @@ if [[ -n "${APP_PROXY}" ]]; then
   cat << EOF > "${WORKBENCH_PROXY_AGENT_SERVICE}"
 [Unit]
 Description=Workbench App Proxy Agent Service
-After=local-fs.target network-online.target
-After=google-guest-agent.service google-startup-scripts.service
-Before=shutdown.target
```
rogerwangcs, line 757: These are no longer needed?

yuhuyoyo: seems like this script is run pretty early on in the boot, it will hang sometimes when the condition doesn't meet.. so I just removed it to be same as vertex

rogerwangcs (approved):

PR #26, opened 2023-11-07
start the custom wsm proxy agent later in the start up script

yuhuyoyo:


rogerwangcs (approved):

startupscript/dataproc/startup.sh
```diff
@@ -1005,6 +973,37 @@ systemctl daemon-reload
 # The jupyter service will be restarted by the default Dataproc startup script
 # and pick up the modified service configuration and environment variables.
 
+###################################
+# Start workbench app proxy agent
```
mbookman, line 977: Can this be changed to:

> Configure and start the workbench app proxy agent

mbookman (changes requested):
Can you add a comment to this PR indicating _why_ the configuration and start needed to be moved later? I'm not sure how to interpret these changes without that insight.

PR #28, opened 2023-11-16
[BENCH-1639] Separate exit_status init from previous command

dexamundsen:
RC - variable exit_code was being set only in the context of the previous statement, and hence not accessible in the subsequent if block
fix - init the exist_status variable in a new line so that it is accessible in rest of code flow

### DIFF
--- a/startupscript/dataproc/startup.sh
+++ b/startupscript/dataproc/startup.sh
@@ -366,8 +366,9 @@ readonly TERRA_SERVER
 # If the server environment is a verily server, use the verily download script.
 if [[ "${TERRA_SERVER}" == *"verily"* ]]; then
   # Map the CLI server to appropriate AFS service path and fetch the CLI distribution path
-  versionJson="$(curl -s "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")" || exit_code=$?
-  if [[ "${exit_code}" -ne 0 ]]; then
+  versionJson="$(curl -s "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")"
+  exit_status=$?
+  if [[ "${exit_status}" -ne 0 ]]; then
       >&2 echo "ERROR: ${TERRA_SERVER} is not a known server"
       exit 1
   fi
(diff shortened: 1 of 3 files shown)


mikenev (approved):

startupscript/dataproc/startup.sh
```diff
@@ -366,8 +366,9 @@ readonly TERRA_SERVER
 # If the server environment is a verily server, use the verily download script.
 if [[ "${TERRA_SERVER}" == *"verily"* ]]; then
   # Map the CLI server to appropriate AFS service path and fetch the CLI distribution path
-  versionJson="$(curl -s "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")" || exit_code=$?
-  if [[ "${exit_code}" -ne 0 ]]; then
+  versionJson="$(curl -s "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")"
+  exit_status=$?
+  if [[ "${exit_status}" -ne 0 ]]; then
```
mbookman, line 371: I don't expect this to actually work, since `set -o errexit` has been enabled. I would expect a curl failure to be a hard failure.

Sorry I didn't catch this earlier. The right way to do this is either:

```
  set +o errexit
  versionJson="$(curl -s "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")"
  exit_status=$?
  set -o errexit
  if [[ "${exit_status}" -ne 0 ]]; then
      >&2 echo "ERROR: ${TERRA_SERVER} is not a known server"
      exit 1
  fi
```

or I think better would be one of:

```
if ! versionJson="$(curl --stderr /tmp/stderr.txt "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")"; then
  >&2 echo "ERROR: Failed to get version file from ${TERRA_SERVER}"
  >&2 cat /tmp/stderr.txt
  exit 1
fi
```

or

```
if ! versionJson="$(curl --no-progress-meter "https://${TERRA_SERVER/verily/terra}-axon.api.verily.com/version")"; then
  >&2 echo "ERROR: Failed to get version file from ${TERRA_SERVER}"
  >&2 echo "${versionJson}"
  exit 1
fi
```

Probably the latter.

PR #29, opened 2023-11-16
[BENCH-1639] simplify error check in AFS call

dexamundsen:
Addressing https://github.com/verily-src/workbench-app-devcontainers/pull/28#discussion_r1396169901


mbookman (approved):
LGTM. Thanks Dex!

PR #30, opened 2023-11-20
Use alternate option for curl

dexamundsen:
Curl does not support --no-progress-bar during terra-cli installation, use alternate -s option
this addresses UI nightly errors


rogerwangcs (approved):

mbookman (approved):
Prefer `--silent`, but fixing the breakage is priority.

Is the lack of `--no-progress-meter` due to old versions of curl?
I had tested this on a JupyterLab cloud environment and it had worked:

```
(base) jupyter@mbb-single-cell-vm:/tmp$ curl --version
curl 7.74.0 (x86_64-pc-linux-gnu)
```

From https://curl.se/docs/manpage.html, I see:

>--no-progress-meter
>...
> Added in 7.67.0.

PR #34, opened 2023-11-24
[BENCH-1051] Support the rename of terra cli to wb

cbookg:


amygdala:
> Code-wise, this all looks very good. I have a few questions:
> 
> * Is there a Jira ticket associated with this, notably which includes sign-off from the Product team on the naming (I know we did a round of voting some time back)?
> * Are there any dependencies on other code or docs that this needs to be coordinated with? I'm wondering whether we should soft link both `~/.terra` -> `~/.vwb` as well as the executable itself such that dependent code (including some solutions getting started material) can transition

There are many docs that would need to be changed (in addition to the examples), so I think we need to alias/soft-link.
(Having not looked at the code changes yet)

cbookg:
https://verily.atlassian.net/browse/BENCH-1051

Soft-linking .terra to .vwb sounds good to me.

dexamundsen (approved):

PR #37, opened 2023-11-29
[BENCH-1686] grant app user access to /user/bin/terra

dexamundsen:


mbookman (commented):
Looks fine, but I'm also curious if granting read access is actually needed.
Elsewhere in the startup scripts we just add execute, since the default umask would already have read enabled. Would prefer consistency across the scripts.

dexamundsen:
> Looks fine, but I'm also curious if granting read access is actually needed. [Elsewhere in the startup scripts](https://github.com/search?q=repo%3Averily-src%2Fworkbench-app-devcontainers%20chmod&type=code) we just add execute, since the default umask would already have read enabled. Would prefer consistency across the scripts.

@yuhuyoyo noticed that users dont have access to /usr/bin hence this is needed. 
Yu, could you also confirm if access to the containing folder is also required to be changed?

yuhuyoyo:
> > Looks fine, but I'm also curious if granting read access is actually needed. [Elsewhere in the startup scripts](https://github.com/search?q=repo%3Averily-src%2Fworkbench-app-devcontainers%20chmod&type=code) we just add execute, since the default umask would already have read enabled. Would prefer consistency across the scripts.
> 
> @yuhuyoyo noticed that users dont have access to /usr/bin hence this is needed. Yu, could you also confirm if access to the containing folder is also required to be changed?
> 
> indeed, app user does not have /usr/bin permission. but also i just tested on a VM that chmod +x terra is enough

yuhuyoyo:
sorry I didn't realize the message was never sent.

@dexamundsen i'd like this to go in before mikey launch

startupscript/post-startup.sh
```diff
@@ -126,7 +126,8 @@ if [[ "${TERRA_SERVER}" == *"verily"* ]]; then
 
   ${RUN_AS_LOGIN_USER} "\
     curl -L https://storage.googleapis.com/${cliDistributionPath#gs://}/download-install.sh | TERRA_CLI_SERVER=${TERRA_SERVER} bash && \
-    cp terra '${TERRA_INSTALL_PATH}'"
+    cp terra '${TERRA_INSTALL_PATH}'
+    chmod +rx '${TERRA_INSTALL_PATH}'"
```
yuhuyoyo, line 130: chmod +x

yuhuyoyo (approved):

startupscript/post-startup.sh
dexamundsen: done

PR #39, opened 2023-12-08
Dexamundsen patch 1

dexamundsen:
Issue: the terra installation runs as login user, and mv terra installation is not needed to be run as that user
fix: split it

fix identified and provided by @yuhuyoyo


dexamundsen:
> Does this change resolve an issue?

Yu gave this as a ref https://github.com/yuhuyoyo/devcontainers/blob/main/post-startup.sh#L126-L127 (which she verified)

PR #43, opened 2024-01-15
BENCH-1933: modularize part of the post-startup.sh

yuhuyoyo:
https://verily.atlassian.net/browse/BENCH-1933

install java as a devcontainer feature.

next step is to pull out gcp specific part into a gcp folder.

### DIFF
--- a/startupscript/bash-completion.sh
+++ b/startupscript/bash-completion.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# bash_completion is installed on Vertex AI notebooks, but the installed
+# completion scripts are *not* sourced from /etc/profile.
+# If we need it system-wide, we can install it there, but otherwise, let's
+# keep changes localized to the user.
+#
+emit "Configuring bash completion for the VM..."
+
+cat << 'EOF' >> "${USER_BASHRC}"
+
+# Source available global bash tab completion scripts
+if [[ -d /etc/bash_completion.d ]]; then
+ for BASH_COMPLETION_SCRIPT in /etc/bash_completion.d/* ; do
+   source "${BASH_COMPLETION_SCRIPT}"
+ done
+fi
+
+# Source available user installed bash tab completion scripts
+if [[ -d ~/.bash_completion.d ]]; then
+ for BASH_COMPLETION_SCRIPT in ~/.bash_completion.d/* ; do
+   source "${BASH_COMPLETION_SCRIPT}"
+ done
+fi
+EOF
--- a/startupscript/bash-completion.sh
+++ b/startupscript/bash-completion.sh
@@ -1,10 +1,15 @@
 #!/bin/bash
 
-# bash_completion is installed on Vertex AI notebooks, but the installed
-# completion scripts are *not* sourced from /etc/profile.
-# If we need it system-wide, we can install it there, but otherwise, let's
-# keep changes localized to the user.
+# bash-completion.sh
 #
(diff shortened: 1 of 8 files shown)


startupscript/bash-completion.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+# bash_completion is installed on Vertex AI notebooks, but the installed
```
mbookman, line 3: Are these startup scripts used on Vertex AI notebooks or Mikey apps? If the latter, then this comment needs to be updated.

startupscript/bash-completion.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
mbookman, line 1: It would be great if each of these smaller files had a header comment which provided a little bit about the context in which they are called. For example:

```
# bash-completion.sh
#
# Writes a block to the user's ~/.bashrc file such that Bash shells pick up
# any Bash completion scripts that have been installed.
#
# Note that this script is intended to be source from the "post-startup.sh" script
# and is dependent on some functions and variables already being set up:
#
# - emit (function)
# - USER_BASHRC: path to user's ~/.bashrc file
```

startupscript/post-startup.sh
```diff
@@ -12,13 +12,12 @@ fi
 
 user="$1"
 workDirectory="$2"
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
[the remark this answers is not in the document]
yuhuyoyo: i get a relative path with that..

this wierd script is based on [this answer ](https://stackoverflow.com/a/3915420)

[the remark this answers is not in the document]
mbookman: Is a relative path a problem? Often preserving the path used by the caller is most helpful.

But if an absolute path is needed, then a comment indicating that and why would be good.
If we do have `readlink` available, then I'd recommend just using `readlink -f` to get an absolute and canonical path.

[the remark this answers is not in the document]
yuhuyoyo: relative link is not reliably becuase we move around directory in the script. for example, we go to /tmp at line 49. i have hesitation with readlink because this script is executed in the container and we don't have control over the app container image. but i agree with you that it' s more readible.

startupscript/bash-completion.sh
yuhuyoyo: updated this one and the other two scripts.

startupscript/emit.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
mbookman, line 1: Please add a comment:

```
# emit.sh
#
# Creates a bash function that can be used for emitting logging messages with a date/timestamp

```

startupscript/bash-completion.sh
```diff
@@ -0,0 +1,30 @@
...
+# bash-completion.sh
+#
+# Writes a block to the user's ~/.bashrc file such that Bash shells pick up
+# any Bash completion scripts that have been installed.
+#
+# Note that this script is intended to be source from the "post-startup.sh" script
+# and is dependent on some functions and variables already being set up:
+#
+# - emit (function)
+# - USER_BASHRC: path to user's ~/.bashrc file
```
mbookman, line 12: Please add a blank line after the comment/before the code

mbookman (approved):
Two small things, but otherwise, LGTM!

PR #48, opened 2024-01-23
local variable use lower case to avoid clashing with the main script

yuhuyoyo:


mbookman (commented):
In this context, you are defining this as a script-local variable (as opposed to a function-local variable)?

Since the script is sourced, then "cloud" becomes available to the parent (the post-startup script) as a side-effect. I think that we should establish some patterns to avoid that.

As this script is always sourced, why pass the parameter instead of just picking up the variable from the environment. Otherwise, we can do things like:

```
declare FOO=...


unset FOO
```

to avoid side effects.

yuhuyoyo:
> In this context, you are defining this as a script-local variable (as opposed to a function-local variable)?
> 
> Since the script is sourced, then "cloud" becomes available to the parent (the post-startup script) as a side-effect. I think that we should establish some patterns to avoid that.
> 
> As this script is always sourced, why pass the parameter instead of just picking up the variable from the environment. Otherwise, we can do things like:
> 
> ```
> declare FOO=...
> 
> 
> unset FOO
> ```
> 
> to avoid side effects.

i renamed user to USER_NAME in the first PR and i forgot to rename user in a subscript and break the script.  so it feels unreliable to depends on the master script to set certain variable.. but i guess i should just add a test to run the script in a github action to catch this.

mbookman:
A few additional patterns to consider:

1- The sub-script header contains a comment indicating the expected variables and functions, this would be one of them
2- We can add a section at the start of the script that will validate that all expected variables are set

```
# Fail fast and consistently if required variables are not set
declare -F emit > /dev/null || (echo "Function emit does not exist" && exit 1)
declare -F get_metadata_value > /dev/null || (echo "Function get_metadata_value does not exist" && exit 1)

echo "${RUN_AS_LOGIN_USER}" >/dev/null
echo "${WORKBENCH_INSTALL_PATH}" >/dev/null
echo "${WORKBENCH_LEGACY_PATH}" >/dev/null
echo "${USER_BASH_COMPLETION_DIR}" >/dev/null
```

yuhuyoyo:
i will apply the fail-fast suggesiton in a follow-up

PR #50, opened 2024-01-31
BENCH-1769: Install latest version of R in dataproc startup script

rogerwangcs:
Adds an initialization action to install the latest version of R on all dataproc cluster nodes. See https://verily.atlassian.net/browse/BENCH-1769 for more details.

For the short term, we need to put this script in our public gcs bucket as initialization scripts only supports public gcs files, so we can't use a github file url.

Followup work: Modify WSM to pass this script as an initialization action on cluster creation.

### DIFF
--- a/startupscript/dataproc/install-r.sh
+++ b/startupscript/dataproc/install-r.sh
@@ -49,7 +49,8 @@ readonly OUTPUT_FILE="${USER_WORKBENCH_CONFIG_DIR}/install-r-output.txt"
 
 readonly CONDA_DIR='/opt/conda/miniconda3'
 readonly CONDA_BIN_DIR="${CONDA_DIR}/bin"
-readonly R_BIN_DIR='/usr/lib/R/bin'
+readonly R_HOME_DIR='/usr/lib/R'
+readonly R_BIN_DIR="${R_HOME_DIR}/bin"
 readonly RUN_R="${R_BIN_DIR}/R"
 
 # Split stdout and stderr in the rest of this script to an output file for debugging.
@@ -101,7 +102,7 @@ if [[ "${DATAPROC_NODE_ROLE}" == 'Master' ]]; then
   cat <<EOF >"${IR_KERNEL}"
 {
   "argv": [
-    "/usr/lib/R/bin/R",
+    "${RUN_R}",
     "--slave",
     "-e",
     "IRkernel::main()",
@@ -111,7 +112,7 @@ if [[ "${DATAPROC_NODE_ROLE}" == 'Master' ]]; then
   "display_name": "R",
   "language": "R",
   "env": {
-    "R_HOME": "/usr/lib/R",
+    "R_HOME": "${R_HOME_DIR}",
     "SPARK_HOME": "/usr/lib/spark"
   }
 }
--- a/startupscript/dataproc/install-r.sh
+++ b/startupscript/dataproc/install-r.sh
@@ -6,33 +6,60 @@
 # This script is intended to be executed as a Dataproc cluster initialization
 # action so such that R is installed and configured on all cluster nodes during
 # creation and autoscaling.
+#
+# This action depends on some user home directory directories being created.
+# Depending on the cluster node type, these directories may not exist. The workbench
+# startup script
+
 # For more information on Initialization actions, see:
 # https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/init-actions
 
-set -o errexit
-set -o nounset
-set -o pipefail
-set -o xtrace
+# Configure shell
+set -o errexit  # Exit on error
+set -o nounset  # Treat unset variables as error
+set -o pipefail # Surface errors inside pipelines
+set -o xtrace   # Output commands before executing them
 
-# Emit a message with a timestamp
+# Utility function to emit a message with a timestamp
 function emit() {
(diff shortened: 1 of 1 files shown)


startupscript/dataproc/install-r.sh
```diff
@@ -0,0 +1,57 @@
...
+# Define directories and system files
+readonly LOGIN_USER="dataproc"
+readonly USER_HOME_DIR="/home/${LOGIN_USER}"
+readonly USER_WORKBENCH_CONFIG_DIR="${USER_HOME_DIR}/.workbench"
+readonly USER_BASHRC="${USER_HOME_DIR}/.bashrc"
+readonly R_BIN_DIR='/usr/lib/R/bin'
+readonly OUTPUT_FILE="${USER_WORKBENCH_CONFIG_DIR}/install-r-output.txt"
+
+# Send stdout and stderr from this script to a file for debugging.
+# Make the .wb directory as the user so that they own it and have correct linux permissions.
```
mbookman, line 31: The second line in this comment doesn't appear to be reflected in code.
(And the directory is `.workbench`, rather than `.wb`.)

startupscript/dataproc/install-r.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
mbookman, line 1: Add a blank line here after the shebang line.

mbookman (changes requested):
LGTM with two small nits.

startupscript/dataproc/install-r.sh
rogerwangcs: Ah, thank you

mbookman (approved):
LGTM

rogerwangcs:
@mbookman @yuhuyoyo 

The R kernel wasn't working right after installing a new version of R. Added changes to:
1. Install R jupyterlab kernel
2. Allow stdout, stderr to stil output so that users can also debug issues via the initialization actions output files
3. Make sure the output file is successfully created in worker nodes too.

startupscript/dataproc/install-r.sh
```diff
@@ -26,6 +26,9 @@ function emit() {
 }
 readonly -f emit
 
+# Retrieve and set the dataproc node type
+readonly ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
```
mbookman, line 30: Not a blocker, but `NODE_ROLE` or `DATAPROC_ROLE` would be preferred here.

mbookman (approved):
LGTM.

Added a suggestion, but it's just a suggestion.

startupscript/dataproc/install-r.sh
```diff
@@ -0,0 +1,123 @@
...
+# Install IRKernel on the master node to support interactive R notebooks in Jupyter
+# Note: We set the PATH to ensure that R knows where to find the 'jupyter' binary
+# See IR kernel installation docs:
+# https://irkernel.github.io/installation/
+if [[ "${DATAPROC_NODE_ROLE}" == 'Master' ]]; then
+  PATH="${CONDA_BIN_DIR}:${PATH}" "${RUN_R}" -e "\
+  install.packages('IRkernel', repos='http://cran.rstudio.com/');
+  IRkernel::installspec()"
+
+  # Point IR jupyter kernel to the newly installed R
```
rogerwangcs, line 99: Another FYI for reviewers: I had to further tweak the R kernel config file itself to ensure the R environment variables point to the latest installed version of R. A general concern is that users may need to be aware that R is installed and managed by the system package manager, which others are managed by conda.

startupscript/dataproc/install-r.sh
```diff
@@ -0,0 +1,127 @@
...
+  "language": "R",
+  "env": {
+    "R_HOME": "${R_HOME_DIR}",
+    "SPARK_HOME": "/usr/lib/spark"
+  }
+}
+EOF
+
+  # Restart jupyter service to pick up the new R kernel
+  readonly JUPYTER_SERVICE_NAME='jupyter.service'
```
mbookman, line 125: This feels like a strange place to declare a readonly variable. Any reason not to declare it up around the other R variables (lines 55-57), like in [post-startup.sh](https://github.com/verily-src/workbench-app-devcontainers/blob/772786274b0d32d8748982f5c6dc7e82891906dc/startupscript/dataproc/startup.sh#L98)?

startupscript/dataproc/install-r.sh
```diff
@@ -0,0 +1,127 @@
...
+set -o xtrace   # Output commands before executing them
+
+# Utility function to emit a message with a timestamp
+function emit() {
+  echo "$(date '+%Y-%m-%d %H:%M:%S') $*"
+}
+readonly -f emit
+
+# Retrieve and set the dataproc node type
+readonly DATAPROC_NODE_ROLE=$(/usr/share/google/get_metadata_value attributes/dataproc-role)
```
mbookman, line 33:
```suggestion
readonly DATAPROC_NODE_ROLE="$(/usr/share/google/get_metadata_value attributes/dataproc-role)"
```

mbookman (changes requested):

startupscript/dataproc/install-r.sh
rogerwangcs: Ah, thanks

mbookman (approved):
LGTM

PR #53, opened 2024-02-14
Install wget

yuhuyoyo:


startupscript/aws/vm-metadata.sh
```diff
@@ -15,6 +15,9 @@ function get_metadata_value() {
     exit 1
   fi
   local tag_key=vwbapp:"$1"
+  
+  apt-get update
```
mbookman, line 19: Do the instances have neither `wget` nor `curl`?

I don't think installing `wget` should be inside the function. Was that an intentional choice? If so, the reasoning should be documented.

It would also be nice to use `--yes` instead of `-y` and to general prefer long options over short.

mbookman (changes requested):

startupscript/aws/vm-metadata.sh
yuhuyoyo: i moved it to the post-startup.sh in the other PR

yuhuyoyo: will drop this one

PR #54, opened 2024-02-14
BENCH-2139: set up custom env variables

yuhuyoyo:
https://verily.atlassian.net/browse/BENCH-2139

### DIFF
--- a/startupscript/post-startup.sh
+++ b/startupscript/post-startup.sh
@@ -57,7 +57,7 @@ exec 2>&1
 
 # The apt package index may not be clean when we run; resynchronize
 apt-get update
-apt install -y jq curl tar
+apt install -y jq curl tar wget
 
 # Create the target directories for installing into the HOME directory
 ${RUN_AS_LOGIN_USER} "mkdir -p '${USER_BASH_COMPLETION_DIR}'"
@@ -71,10 +71,17 @@ if [[ -e ~/.bashrc ]]; then
 fi
 
 EOF
-##################################################
-# Set up user bashrc with workbench customization
-##################################################
-source ${SCRIPT_DIR}/setup-bashrc.sh
+
+# Indicate the start of Workbench customizations of the ~/.bashrc
+cat << EOF >> "${USER_BASHRC}"
+### BEGIN: Workbench-specific customizations ###
+
+# Prepend "/usr/bin" (if not already in the path)
+if [[ "${PATH}:" != "/usr/bin:"* ]]; then
+  export PATH=/usr/bin:${PATH}
+fi
+
+EOF
 
 ##################################################
 # Set up java which is required for workbench CLI 
@@ -85,6 +92,11 @@ source ${SCRIPT_DIR}/install-java.sh
 # Install workbench CLI
 ###################################
 source ${SCRIPT_DIR}/install-cli.sh 
+
+##################################################
+# Set up user bashrc with workbench customization
+##################################################
+source ${SCRIPT_DIR}/setup-bashrc.sh
 
 #################
 # bash completion
--- a/startupscript/post-startup.sh
+++ b/startupscript/post-startup.sh
@@ -71,15 +71,10 @@ if [[ -e ~/.bashrc ]]; then
 fi
 
 EOF
-
-# Indicate the start of Workbench customizations of the ~/.bashrc
-cat << EOF >> "${USER_BASHRC}"
-
-# Prepend "/usr/bin" (if not already in the path)
-if [[ "${PATH}:" != "/usr/bin:"* ]]; then
-  export PATH=/usr/bin:${PATH}
-fi
-EOF
+##################################################
+# Set up user bashrc with workbench customization
+##################################################
+source ${SCRIPT_DIR}/setup-bashrc.sh
 
 ##################################################
 # Set up java which is required for workbench CLI 
--- a/startupscript/setup-bashrc.sh
+++ b/startupscript/setup-bashrc.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+# setup-bashrc.sh
+#
+# Set up customizaiton for user bashrc file.
+#
+# Note that this script is intended to be source from the "post-startup.sh" script 
+# and is dependent on some variables already being set up:
+#
+# - emit (function)
+# - USER_BASHRC
+# - CLOUD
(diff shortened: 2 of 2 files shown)


startupscript/post-startup.sh
```diff
@@ -91,6 +92,11 @@ source ${SCRIPT_DIR}/install-java.sh
 ###################################
 source ${SCRIPT_DIR}/install-cli.sh 
 
+##################################################
+# Set up user bashrc with workbench customization
+##################################################
+source ${SCRIPT_DIR}/setup-bashrc.sh
```
mbookman, line 98:
```suggestion
source "${SCRIPT_DIR}/setup-bashrc.sh"
```

startupscript/post-startup.sh
```diff
@@ -91,6 +92,11 @@ source ${SCRIPT_DIR}/install-java.sh
 ###################################
 source ${SCRIPT_DIR}/install-cli.sh
```
mbookman, line 93:
```suggestion
source "${SCRIPT_DIR}/install-cli.sh"
```

startupscript/setup-bashrc.sh
```diff
@@ -0,0 +1,61 @@
...
+# Set up customizaiton for user bashrc file.
+#
+# Note that this script is intended to be source from the "post-startup.sh" script 
+# and is dependent on some variables and packages already being set up:
+#
+# - emit (function)
+# - USER_BASHRC: path to user's ~/.bashrc file
+# - CLOUD: aws/gcp
+# - LOG_IN: whether the user is logged into the wb CLI as part of the script
+# - RUN_AS_LOGIN_USER: run command as user
```
mbookman, line 13:
```suggestion
# - RUN_AS_LOGIN_USER: run command as non-root Unix user (ex: jupyter, dataproc)
```

startupscript/post-startup.sh
mbookman: In theory the SCRIPT_DIR can contain spaces, so it should always be quoted.

startupscript/setup-bashrc.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
mbookman, line 1:
```suggestion
#!/bin/bash

# setup-bashrc.sh
```

startupscript/setup-bashrc.sh
```diff
@@ -0,0 +1,61 @@
...
+# setup-bashrc.sh
+#
+# Set up customizaiton for user bashrc file.
```
mbookman, line 4:
```suggestion
# Set up customization for user bashrc file.
```

startupscript/setup-bashrc.sh
mbookman, line 1: It would be great if we could pull in all of the commentary from:

https://github.com/verily-src/mc-terra-authz-wsm/blob/60048df8e901c6bf35b3be96fa0969380ba15236/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh#L128

or we set somewhere else as the "canonical" documentation around these environment variables. Currently it says:

```
# Set variables into the .bash_profile such that they are available
# to terminals, notebooks, and other tools
#
# We have new-style variables (eg GOOGLE_CLOUD_PROJECT) which are set here
# and CLI (terra app execute env).
# We also support a few variables set by Leonardo (eg GOOGLE_PROJECT).
# Those are only set here and NOT in the CLI as they are intended just
# to make porting existing notebooks easier.

# Keep in sync with terra CLI environment variables:
# https://github.com/DataBiosphere/terra-cli/blob/14cf51dd809573c7ae9a3ef10ddd427fa057cb8f/src/main/java/bio/terra/cli/app/CommandRunner.java#L88

# *** Variables that are set by Leonardo for Cloud Environments
# (https://github.com/DataBiosphere/leonardo)
```

The pointer to the terra CLI repository should be updated as well.

startupscript/setup-bashrc.sh
```diff
@@ -0,0 +1,61 @@
...
+
+  # PET_SA_EMAIL is the pet service account for the Workbench user and
+  # is specific to the GCP project backing the workspace
+  readonly PET_SA_EMAIL="$(
+    ${RUN_AS_LOGIN_USER} "wb auth status --format=json" | \
+    jq --raw-output ".serviceAccountEmail")"
+
+  cat << EOF >> "${USER_BASHRC}"
+# Set up a few legacy Workbench-specific convenience variables
+export TERRA_USER_EMAIL='${OWNER_EMAIL}'
```
mbookman, line 32: As this reads, both `TERRA_USER_EMAIL` and `WORKBENCH_USER_EMAIL` are "legacy" convenience variable. I think that's not what was intended.

mbookman (changes requested):

startupscript/setup-bashrc.sh
```diff
@@ -0,0 +1,71 @@
...
+# We also support a few variables set for legacy reasons (eg GOOGLE_PROJECT). Those are only set here and 
+# NOT in the CLI as they are intended just to make porting existing notebooks easier.
+# 
+# Keep in sync with Workbench CLI environment variables:
+# https://github.com/verily-src/terra-tool-cli/blob/b146951ffc9c4f72f4d9c491a543b5c29bea3650/src/main/java/bio/terra/cli/app/CommandRunner.java#L94
+#
+# Note that this script is intended to be source from the "post-startup.sh" script 
+# and is dependent on some variables and packages already being set up:
+#
+# - emit (function)
```
mbookman, line 16:
```suggestion
# - emit:  function to echo a message with a timestamp
```

startupscript/setup-bashrc.sh
```diff
@@ -0,0 +1,71 @@
...
+# - USER_BASHRC: path to user's ~/.bashrc file
+# - CLOUD: aws/gcp
+# - LOG_IN: whether the user is logged into the wb CLI as part of the script
+# - RUN_AS_LOGIN_USER: run command as non-root Unix user (ex: jupyter, dataproc)
+# - must be run after install-cli.sh
```
mbookman, line 21:
```suggestion
# 
# This script must be run after install-cli.sh
```

mbookman (approved):
Some small things, but otherwise LGTM!

PR #69, opened 2024-03-18
scan mounted buckets folders for *.zarr folder and create cirro dataset

yuhuyoyo:


src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# cirro_scanner.sh scans the mounted buckets to find cirro data. Creates a cirro dataset when *.zarr folder
+# is found and is not yet in the existing cirro datasets.
+
```
mbookman, line 5:
```suggestion
set -o errexit
set -o nounset
set -o pipefail
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+if [[ $# -ne 1 ]]; then
+  echo "Usage: $0 <true/false>"
```
mbookman, line 7:
```suggestion
  echo "Usage: $0 <true|false>"
```

mbookman: Please add a usage description to the top level script comment such that the caller does not need to read the code below to figure out what true|false represent.

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+  exit 1
+fi
+
+readonly LOG_IN="${1}"
+if [[ "${LOG_IN}" == "false" ]]; then
```
mbookman, line 12: This is an unusual usage pattern. I don't really think that the script should exit with a clean exit code (0) if it didn't do anything.

If the user is not logged in, then does "wb resource mount" fail in a way that makes it clear? Alternatively, it feels simpler to just document that this script should only be called when the user is already logged in to Workbench (clearly the caller already knows). If there's a common usage where the script is called without a user logged in already _and_ that usage needs to exit cleanly (exit 0), as the behavior of this script is unexpected.

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+  echo "user is not logged in"
+  exit 0
+fi
+
+wb resource mount
+
+# Define the function to check if folder path exists in the JSON array
+dataset_exists() {
```
mbookman, line 20:
```suggestion
function dataset_exists() {
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+    local url="$1"
+    local array_json=$(curl -s localhost:3000/api/datasets)
```
mbookman, line 22:
```suggestion
    local array_json="$(curl -s localhost:3000/api/datasets)"
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+    local exists=$(echo "$array_json" | jq -e --arg path "$url" '.[] | select(.url == $path)' | wc -l)
```
mbookman, line 23: Why set the variable just to echo it? How about replacing these two lines with:

```
echo "$array_json" | \
  jq -e --arg path "$url" '.[] | select(.url == $path)' | \
  wc -l
```

The function should be renamed and have a comment indicating that it outputs a numerical value. "dataset_exists" implies a boolean return.

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+    echo "$exists"
+}
+readonly -f dataset_exists
+
+# Create a cirro dataset
+create_dataset() {
```
mbookman, line 29:
```suggestion
function create_dataset() {
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+    local url="$1"
+    local name=$(basename "$url")
```
mbookman, line 31:
```suggestion
    local name="$(basename "${url}")"
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+    curl -X POST -F "name=$name" -F "url=$url" localhost:3000/api/dataset
```
mbookman, line 32:
```suggestion
    curl -X POST -F "name=${name}" -F "url=${url}" localhost:3000/api/dataset
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+}
+readonly -f create_dataset
+
+# Define the function to perform the scanning
+scan_folders_and_create_datasets() {
```
mbookman, line 37:
```suggestion
function scan_folders_and_create_datasets() {
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+    # Search for folders with name *.zarr and create cirro dataset if it doesn't exist
+    find /root/workspace -type d -name '*.zarr' | while read -r folder; do
+        if [[ "$(dataset_exists "$folder")" -eq "0" ]]; then
```
mbookman, line 40:
```suggestion
        if [[ "$(dataset_exists "${folder}")" -eq "0" ]]; then
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,57 @@
...
+            create_dataset "$folder"
+        else
+            echo "Folder '$folder' already exists in the cirro server dataset."
```
mbookman, line 43:
```suggestion
            echo "Folder '${folder}' already exists in the cirrocumulus server dataset."
```

mbookman (changes requested):

src/cirrocumulus/scripts/cirro_scanner.sh
yuhuyoyo: ok i will document it then

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,52 @@
...
+set -o nounset
+set -o pipefail
+
+wb resource mount
+
+# Define the function to check for existing cirrocumulus datasets with the given path.
+function matching_dataset_count() {
+    local url="$1"
+    curl -s localhost:3000/api/datasets | \
+        jq -e --arg path "$url" '.[] | select(.url == $path)' | \
```
mbookman, line 17:
```suggestion
        jq -e --arg path "${url}" '.[] | select(.url == ${path})' | \
```

src/cirrocumulus/scripts/cirro_scanner.sh
```diff
@@ -0,0 +1,51 @@
...
+        if [[ "$(matching_dataset_count "${folder}")" -eq "0" ]]; then
+            create_dataset "${folder}"
+        else
+            echo "Folder '${folder}' already exists in the cirrocumulus server dataset."
+        fi
+    done
+}
+readonly -f scan_folders_and_create_datasets
+
+wb resource mount
```
mbookman, line 42:
```suggestion
wb resource mount

```

mbookman (approved):
One small nit (missing blank line).
Otherwise LGTM.

PR #70, opened 2024-03-19
retry nextflow install

yuhuyoyo:
add retry to the nextflow install command
https://verily.atlassian.net/browse/BENCH-1935


startupscript/dataproc/startup.sh
```diff
@@ -211,6 +211,47 @@ function exit_handler {
...
+  # If the exit code is non-zero (i.e. command failed), and we have not
+  # reached the maximum number of retries, run the command again
+  if [[ ${exit_code} -ne 0 && ${retries} -gt 0 ]]; then
+    retry $((retries - 1)) "${command}"
+  else
+    # Return the exit code from the command
+    return ${exit_code}
+  fi
+}
+readonly -f retry
```
mbookman, line 245: Would recommend something more compact like this modified version of code from `dsub`:

http://google3/third_party/py/dsub/dsub/providers/google_utils.py;l=116;rcl=517278768

```
#######################################
# function to retry a command
#######################################
function retry () {
  local max_attempts="$1"
  local command="$2"

  local attempt
  for ((attempt = 1; attempt <= max_attempts; attempt++)); do
    # Run the command and return if success
    if ${command}; then
      return
    fi

    # Sleep a bit in case the problem is a transient network/server issue
    if ((attempt < max_attempts)); then
      echo "Retrying `${command}` in 5 seconds"
      sleep 5
    fi
 done

 # Execute without the if/then protection such that the exit code propagates
 ${command}
}
readonly -f retry
```

mbookman (changes requested):

mbookman (approved):
LGTM!

PR #73, opened 2024-03-21
Enable Building Workbench Java Code in VS Code

jmczerk:
I'm not sure what the intent of our vscode app is, but these changes are necessary to be able to build Workbench Java code in the app (lock to Java 17, `gcloud`needed for private artifacts), which I have found very useful for testing in-EC2 behavior.

### DIFF
--- a/src/cirrocumulus/.devcontainer.json
+++ b/src/cirrocumulus/.devcontainer.json
@@ -9,6 +9,7 @@
   "postStartCommand": "nohup bash -c '[[ ${templateOption:login} == true ]] && ./scripts/cirro_scanner.sh &'",
   "features": {
       "ghcr.io/devcontainers/features/aws-cli:1": {},
+      "ghcr.io/dhoeric/features/google-cloud-cli:1": {},
       "ghcr.io/devcontainers/features/java:1": {}
   },
   "remoteUser": "root"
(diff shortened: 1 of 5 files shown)


yuhuyoyo (approved):
LGTM. do you use gcloud to test ec2?

yuhuyoyo:
also can we add this to rstudio too?

mbookman:
Can you clarify what "Workbench Java" is and what we need it for? Is it just for the `wb` CLI or are there other apps we are now deploying on VMs?

jmczerk:
> LGTM. do you use gcloud to test ec2?

No, it is needed for gradle to pull artifacts from the private GAR that `verily-src` forks publish to now.  Specifically one needs to `gcloud auth login` to be able to run any gradle builds.

jmczerk:
> Can you clarify what "Workbench Java" is and what we need it for? Is it just for the `wb` CLI or are there other apps we are now deploying on VMs?

Sorry, this should have been "Workbench Java ***code***"... i.e. I want to be able to build and test our product java code from a VM.  Generally I'd expect this to be the CLI, but I could see wanting to run some service integration tests inside an EC2 instance.

This is a really more of a Workbench-internal use case, but I'm not sure I see any other VS Code use cases, so seemed like a reasonable place to put it?

yuhuyoyo:
> > also can we add this to rstudio too?
> 
> Lock Java version there too? Or just add the `gcloud` install?

gcloud

PR #74, opened 2024-03-25
new devcontainer for idle checking

yuhuyoyo:
https://verily.atlassian.net/browse/BENCH-2520

### DIFF
--- a/src/idlechecker/idle-checker.sh
+++ b/src/idlechecker/idle-checker.sh
@@ -7,6 +7,7 @@
 #
 # - AWS CLI is installed
 # - CLOUD: cloud platform VM is running on
+# - bc https://manpages.ubuntu.com/manpages/trusty/en/man1/bc.1.html
 
 set -o errexit
 set -o nounset
@@ -35,7 +36,7 @@ function create_tag() {
 readonly -f create_tag
 
 function set_cpu_last_active() {
-  now = $(date +'%s')
+  now=$(date +'%s')
   if [[ "${CLOUD}" == "gcp" ]]; then
     set_guest_attributes "${now}"
   else
--- a/src/idlechecker/idle-checker.sh
+++ b/src/idlechecker/idle-checker.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# idle-checker.sh checks for cpu utilization of the VM and update last active timestamp if the CPU utilization
+# is above a given threshold.
+
+# threshold - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
+
+readonly threshold="${THRESHOLD:-0.1}"
+
+function set_cpu_last_active() {
+  curl -s -X PUT --data "${attr_value}" \
+    -H "Metadata-Flavor: Google" \
+    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/cpu-utilization/last-active"
+}
+readonly -f set_cpu_last_active
+
+while true
+do
+
+  load=$(uptime | sed -e 's/.*load average: //g' | awk '{ print $1 }') # 1-minute average load
+  load="${load//,}" # remove trailing comma
+  echo "cpu load is $load"
+  res=$(echo $load'<'$threshold | bc -l)
+  if (( $load < $threshold )); then
+    echo "Idling.."
+  else
+    set_cpu_last_active $(date +'%s')
+  fi
+
+  sleep 60
+
+done
--- a/src/idlechecker/idle-checker.sh
+++ b/src/idlechecker/idle-checker.sh
@@ -2,21 +2,18 @@
 
 # idle-checker.sh checks for cpu utilization of the VM and update last active timestamp if the CPU utilization
 # is above a given threshold.
-
-# threshold - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
+#
+# Note that this script is dependent on some variables and packages being set up:
+#
+# - AWS CLI is installed
+# - CLOUD: cloud platform VM is running on
 
 set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
 
-if [[ $# -ne 1 ]]; then
-  echo "Usage: $0 <gcp/aws>"
-  exit 1
-fi
-
-readonly CLOUD="${1}"
-
+# threshold - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
 readonly threshold="${THRESHOLD:-0.1}"
 readonly LAST_ACTIVE_KEY="cpu-utilization/last-active"
 
--- a/src/idlechecker/idle-checker.sh
+++ b/src/idlechecker/idle-checker.sh
@@ -11,14 +11,12 @@
 # Software:
 # - AWS CLI
 
-# - CLOUD: cloud platform VM is running on
-
 set -o errexit
 set -o nounset
 set -o pipefail
 set -o xtrace
 
-# threshold - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
+# THRESHOLD- threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
 readonly THRESHOLD="${THRESHOLD:-0.1}"
 readonly LAST_ACTIVE_KEY="last-active/cpu"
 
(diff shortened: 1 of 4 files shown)


src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+# idle-checker.sh checks for cpu utilization of the VM and update last active timestamp if the CPU utilization
```
mbookman, line 3:
```suggestion
# idle-checker.sh checks for CPU utilization of the VM and updates last active timestamp if the CPU utilization
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+# is above a given threshold.
+#
+# Note that this script is dependent on some variables and packages being set up:
```
mbookman, line 6:
```suggestion
# Note that this script is dependent on some environment variables and packages being set up:
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+#
+# - AWS CLI is installed
```
mbookman, line 8:
```suggestion
# Environment variables:
# - CLOUD: cloud platform VM is running on
#
# Software:
# - AWS CLI
# - bc https://manpages.ubuntu.com/manpages/trusty/en/man1/bc.1.html

```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+# - CLOUD: cloud platform VM is running on
+# - bc https://manpages.ubuntu.com/manpages/trusty/en/man1/bc.1.html
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+# threshold - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
+readonly threshold="${THRESHOLD:-0.1}"
```
mbookman, line 18: You don't *need* to change the name. If you really want to, then make it THRESHOLD_VAL or some such, but this is fine:

```suggestion
readonly THRESHOLD="${THRESHOLD:-0.1}"
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+  echo "Setting cpu-utilization/last-active"
+  curl -s -X PUT --data "$1" \
+    -H "Metadata-Flavor: Google" \
+    "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/${LAST_ACTIVE_KEY}"
+}
+readonly -f set_guest_attributes
+
+function create_tag() {
+  echo "Creating tag vwbapp:cpu-utilization/last-active"
+  INSTANCE_ID="$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)"
```
mbookman, line 31:
```suggestion
  local INSTANCE_ID
  INSTANCE_ID="$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)"
```

I also have a preference for local variables to be lower-cased, so even better would be:

```suggestion
  local id
  id="$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)"
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+  aws ec2 create-tags \
+    --resources "${INSTANCE_ID}" \
+    --tags Key=vwbapp:${LAST_ACTIVE_KEY},Value="$1"
+}
+readonly -f create_tag
+
+function set_cpu_last_active() {
+  now=$(date +'%s')
```
mbookman, line 39:
```suggestion
  local now
  now="$(date +'%s')"
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+  if [[ "${CLOUD}" == "gcp" ]]; then
+    set_guest_attributes "${now}"
+  else
```
mbookman, line 42: Is this a broadly generic, cross platform else condition, or is it AWS specific? If so, then we should:

```
if [[ "${CLOUD}" == "gcp" ]]; then
    set_guest_attributes "${now}"
elif [[ "${CLOUD}" == "aws" ]]; then
    create_tag "${now}"
else
   1>&2 echo "Unexpected cloud platform ${CLOUD}"
   exit 1
fi
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+    create_tag "${now}"
+  fi  
+}
+readonly -f set_cpu_last_active
+
+while true
```
mbookman, line 48: declare LOAD

```suggestion
while true; do
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+do
+
+  load=$(uptime | sed -e 's/.*load average: //g' | awk '{ print $1 }') # 1-minute average load
```
mbookman, line 51:
```suggestion
  LOAD="$(uptime | sed -e 's/.*load average: //g' | awk '{ print $1 }')" # 1-minute average load
```
However, if we know that this is Linux, then I think we should use `/proc/loadavg`:

```
$ cat /proc/loadavg
0.24 0.13 0.04 1/210 3326202
```

hence:

```
LOAD="$(cat /proc/loadavg | awk '{print $1}')"
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+  load="${load//,}" # remove trailing comma
+  echo "cpu load is $load"
```
mbookman, line 53: Feels like it would be great to have this timestamped since we are going to be shutting VMs down in response to time-based observations. Recommend pulling in the "emit" function.

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,62 @@
...
+  if (( $(echo "$threshold > $load" |bc -l) )); then
```
mbookman, line 54: You could get rid of the dependence on `bc` by using `awk`:

```suggestion
  # Check if the LOAD has exceeded the THRESHOLD.
  # Note the use of awk for comparison of real numbers.
  if echo "${THRESHOLD}" "${LOAD}" | awk '{if ($1 > $2) exit 0; else exit 1}'; then
```

mbookman (changes requested):

src/idlechecker/idle-checker.sh
yuhuyoyo: the set guest attributes and create tag functions put the timestamp into the vm metadata. but i can also pull in the emit function here.

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,74 @@
...
+#
+# Note that this script is dependent on some environment variables and packages being set up:
+#
+# Environment variables:
+# - CLOUD: cloud platform VM is running on
+#
+# Software:
+# - AWS CLI
+
+# - CLOUD: cloud platform VM is running on
```
mbookman, line 14: Can be cut

```suggestion
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,74 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+# threshold - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
```
mbookman, line 21:
```suggestion
# THRESHOLD - threshold for cpu usage, used to determine if instance is idle. If usage goes above this number count resets to zero. By default 0.1 (10 percent)
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,74 @@
...
+  id="$(wget -q -O - http://169.254.169.254/latest/meta-data/instance-id)"
+  aws ec2 create-tags \
+    --resources "${id}" \
+    --tags Key=vwbapp:${LAST_ACTIVE_KEY},Value="$1"
+}
+readonly -f create_tag
+
+function set_cpu_last_active() {
+  local now
+  now=$(date +'%s')
```
mbookman, line 50:
```suggestion
  now="$(date +'%s')"
```

src/idlechecker/idle-checker.sh
```diff
@@ -0,0 +1,74 @@
...
+    1>&2 emit "Unexpected cloud platform ${CLOUD}"  
+    exit 1
+  fi  
+}
+readonly -f set_cpu_last_active
+
+declare LOAD
+while true; do
+  LOAD="$(awk '{print $1}' /proc/loadavg)" # 1-minute average load
+  emit "cpu load is ${LOAD}"
```
mbookman, line 65:
```suggestion
  emit "CPU load is ${LOAD}"
  
```

mbookman (changes requested):
Just a few small things...

mbookman (approved):
LGTM!

PR #88, opened 2024-04-04
add a script to shutdown VM

yuhuyoyo:


startupscript/butane/aws/metadata-utils.sh
```diff
@@ -17,16 +17,27 @@ function get_metadata_value() {
...
+# guest attributes are not supported on EC2 instances. But to keep the interface consistent with GCP, we define a no-op function.
+function get_guest_attribute() {
+  get_metadata_value "$1"
+}
+readonly -f get_guest_attribute
+
+
 # Sets tags on the EC2 instance with the given key and value.
 function set_metadata() {
+  if [[ -z "$2" ]]; then
```
mbookman, line 37: What is the bad behavior if the value is empty?
The usage message on its own doesn't indicate that the value cannot be an empty string.

startupscript/butane/idle-shutdown.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+readonly LAST_ACTIVE
+echo "Last active time: ${LAST_ACTIVE}"
+
+NOW="$(date +'%s')"
+readonly NOW
+
+# Check if the VM has been idle for longer than the timeout.
+if [[ $((NOW - LAST_ACTIVE)) -gt IDLE_TIMEOUT_SECONDS ]]; then
+    echo "Shutting down the VM due to inactivity."
```
mbookman, line 36: It would be great to emit NOW, LAST_ACTIVE, and IDLE_TIMEOUT_SECONDS in this message.
Even better if the message itself was timestamped (as is done in the emit function).

mbookman (changes requested):

startupscript/butane/aws/metadata-utils.sh
yuhuyoyo: actually you are right.. it can be empty

startupscript/butane/idle-shutdown.sh
yuhuyoyo: added emit function

startupscript/butane/idle-shutdown.sh
```diff
@@ -0,0 +1,44 @@
...
+fi
+readonly LAST_ACTIVE
+emit "Last active time: ${LAST_ACTIVE}"
+
+NOW="$(date +'%s')"
+readonly NOW
+
+# Check if the VM has been idle for longer than the timeout.
+if [[ $((NOW - LAST_ACTIVE)) -gt IDLE_TIMEOUT_SECONDS ]]; then
+    emit "Shutting down the VM. Last active time: ${LAST_ACTIVE}. Idle timeout threshold is: ${IDLE_TIMEOUT_SECONDS}"
```
mbookman, line 40: I think it is still worth including NOW in the message, since it's a numeric value and helps the reader check the calculation.

mbookman (approved):
One small thing, but otherwise LGTM.

mbookman (approved):
LGTM

PR #89, opened 2024-04-05
[BENCH-855] Allow users to toggle display of hidden files in jupyter browser

zloery:
This change allows users to toggle whether hidden files (prefixed with `.`) are shown in the Jupyter file browser or not. By enabling this setting, Jupyter adds a "Show hidden files" option under the `View` menu. By default this is off, which matches previous behavior.

I've added this for all notebooks and dataproc clusters (including ones via custom image) as this option was [added in Jupyterlab 3.2.0](https://github.com/jupyterlab/jupyterlab/issues/2049)(early 2022). My assumption is that users are not using significantly older versions of Jupyterlab - please let me know if that's not true and I can avoid setting it for custom images.

### DIFF
--- a/startupscript/dataproc/startup.sh
+++ b/startupscript/dataproc/startup.sh
@@ -1078,6 +1078,9 @@ a#workspace {\
 
 ### BEGIN: Workbench-specific customizations ###
 
+# Allow users to toggle display of hidden files in file browser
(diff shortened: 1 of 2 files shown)


wnojopra (approved):
LGTM

yuhuyoyo (approved):

PR #90, opened 2024-04-08
fix aws metadata util script

yuhuyoyo:
should not echo logs in get_metadata_value

### DIFF
--- a/startupscript/butane/aws/metadata-utils.sh
+++ b/startupscript/butane/aws/metadata-utils.sh
@@ -4,7 +4,7 @@
 # to retrieve or modify instance tags. It is run on the VM host. AWS CLI is not installed on Flatcar VM by default so
 # we are running the AWS CLI in a container.
 
-# Retrieves an instance tag set on the VM. If the tag is not set, it returns 0.
+# Retrieves an instance tag set on the VM. If the tag is not set, it returns an empty string..
 function get_metadata_value() {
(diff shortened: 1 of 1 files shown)


startupscript/butane/aws/metadata-utils.sh
```diff
@@ -4,24 +4,29 @@
 # to retrieve or modify instance tags. It is run on the VM host. AWS CLI is not installed on Flatcar VM by default so
 # we are running the AWS CLI in a container.
 
-# Retrieves an instance tag set on the VM.
+# Retrieves an instance tag set on the VM. If the tag is not set, it returns 0.
```
mbookman, line 7: Does returning zero generalize for all such tags?
Why not just return an empty string? That feels more natural.
Alternatively, the function could take an optional second argument which would be the default value to return.

```
MY_STRING="$(get_metadata_value "my-string-value" "")"
MY_NUMBER="$(get_metadata_value "my-numeric-value" "0")"
```

yuhuyoyo: right i got tunnel visioned.. echo empty string instead

mbookman, line 7:
```suggestion
# Retrieves an instance tag set on the VM. If the tag is not set, it returns an empty string..
```

mbookman (approved):
One small thing nit. Otherwise LGTM.

PR #91, opened 2024-04-08
allow user override

yuhuyoyo:


startupscript/butane/idle-shutdown.sh
```diff
@@ -18,12 +19,16 @@ readonly -f emit
...
+readonly IDLE_TIMEOUT_SECONDS
+if [[ -z "${IDLE_TIMEOUT_SECONDS}" ]]; then
+    emit "No idle timeout seconds set. Do not autostop VM."
+    exit 0
+fi
 
 # Get the last time the VM was active.
-LAST_ACTIVE="$(get_guest_attribute "last-active/cpu" || echo "0")"
-LAST_ACTIVE_PROXY="$(get_guest_attribute "last-active/proxy" || echo "0")"
+LAST_ACTIVE="$(get_guest_attribute "last-active/cpu")"
```
cbookg, line 30: this used to be 0 when not set, and now its "".  seems like maybe that will break the compare below?

yuhuyoyo: actually <some value> -lt "" return false and doesn't throw error

startupscript/butane/gcp/metadata-utils.sh
```diff
@@ -12,11 +12,11 @@ function get_metadata_value() {
   local metadata_path="${1}"
   curl --retry 5 -s -f \
     -H "Metadata-Flavor: Google" \
-    "http://metadata/computeMetadata/v1/instance/attributes/${metadata_path}"
+    "http://metadata/computeMetadata/v1/instance/attributes/${metadata_path}" || echo ""
```
yuhuyoyo, line 15: to be consistent with aws behavior

cbookg (approved):

startupscript/butane/idle-shutdown.sh
mbookman: It would be better not to depend on unintuitive or infrequently exercised behavior like: `value1 -lt ""`.

It would be better to have the caller explicitly reinterpret the empty string to 0.
So either having the caller pass a default to get_metadata_value() or doing something like:

```
LAST_ACTIVE="${LAST_ACTIVE:-0}"
LAST_ACTIVE_PROXY="${LAST_ACTIVE_PROXY:-0}"
```

to set the values to 0 if they are empty.

mbookman (changes requested):

startupscript/butane/idle-shutdown.sh
yuhuyoyo: added default value.

startupscript/butane/gcp/metadata-utils.sh
```diff
@@ -3,29 +3,29 @@
 # metadata-utils.sh defines functions for gcp instance attributes and guest attributes. This script is intended to be sourced from other scripts
 # to retrieve or modify instance attributes. It is run on the VM host.
 
-# Retrieves an instance attributes set on the VM.
+# Retrieves an instance attributes set on the VM. If the attribute is not set, it returns the default value.
 function get_metadata_value() {
-  if [[ -z "$1" ]]; then
-    echo "usage: get_metadata_value <tag>"
+  if [[ $# -lt 2 ]]; then
```
mbookman, line 8: FWIW, I'm not opposed to the default value being optional (with "" being the default default).
But let's see how we like requiring the caller to specify the default.

mbookman (approved):
One comment, but otherwise LGTM.

PR #92, opened 2024-04-09
Check uptime first to handle reboot

yuhuyoyo:
on system reboot, we could still have super old timestamp. so check uptime first. if the machine has been up exceeding the time limit, then use the last active timestamp.

### DIFF
--- a/startupscript/butane/idle-shutdown.sh
+++ b/startupscript/butane/idle-shutdown.sh
@@ -29,8 +29,8 @@ if [[ -n "${USER_TIMEOUT_OVERRIDE}" ]]; then
 fi
 
 # Get the last time the VM was active.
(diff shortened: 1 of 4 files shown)


startupscript/butane/idle-shutdown.sh
```diff
@@ -26,6 +26,13 @@ if [[ -z "${IDLE_TIMEOUT_SECONDS}" ]]; then
     exit 0
 fi
 
+UP_TIME="$(awk '{print $1}' /proc/uptime)"
```
mbookman, line 29: Worth adding a comment about the issue of old timestamps on boot here.
It is somewhat unrelated to this check here, but I could see a future contributor thinking that this check was unnecessary and removing it; then not discovering it right away as it only crops up on reboot.

PR #93, opened 2024-04-16
add volume for the home directory to persist data 

yuhuyoyo:


src/shiny/docker-compose.yaml
```diff
@@ -0,0 +1,31 @@
+version: '2.4'
+services:
+  app:
+    container_name: "application-server"
+    build:
+      context: .
+      dockerfile: Dockerfile
+    args:
+      - VARIANT=4
```
mbookman, line 9: Can you add a comment to the docker compose YAML file, or is there some place to document this configuration. How would someone maintaining this know what VARIANT 4 is and why it is set here?

mbookman (changes requested):

src/shiny/docker-compose.yaml
yuhuyoyo: added a comment.

i wanted to make image tag as a templatable variable so user can pick something other than the default. same for other containers. but publishing a devcontainer template to ghcr.io requires a launch and I haven't gotten around to do it.

src/shiny/docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
...
+services:
+  app:
+    container_name: "application-server"
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        - BASE_IMAGE=rocker/shiny  
+        # version tag of the rocker/shiny image.
+        - VARIANT=4
```
mbookman, line 11: A few questions on this. Why not update the Dockerfile and have `BASE_IMAGE` include the tag:

```
BASE_IMAGE=rocker/shiny:latest
```
If there's a compelling reason not to, then change this from `VARIANT` to `IMAGE_TAG` and the comment is not needed.

mbookman (changes requested):

mbookman (approved):
LGTM!

PR #96, opened 2024-05-03
use vwbusr for user set tag

yuhuyoyo:


startupscript/butane/aws/metadata-utils.sh
```diff
@@ -5,12 +5,12 @@
 # we are running the AWS CLI in a container.
 
 # Retrieves an instance tag set on the VM. If the tag is not set, it returns the default vaule.
-function get_metadata_value() {
-  if [[ $# -lt 2 ]]; then
-    echo "usage: get_metadata_value <tag> <default-value>"
+function get_tag() {
+  if [[ $# -lt 3 ]]; then
```
jmczerk, line 9: Nit: usage should also say `get_tag`.

PR #98, opened 2024-05-15
AWS Startup Script Fixes

jmczerk:
Makes the following changes:
- Fixes wrong tag prefix in AWS startup script `get_metadata_value` function (`s/vwbapp/vwbuser`).
- Uses `WorkspaceId` tag for "no-login" path, as this will always be present, and always be set to the Workspace UUID.  Renamed variables `s/WORKBENCH_WORKSPACE/WORKBENCH_WORKSPACE_UUID` to make this explicit.

Note that the "no-login" paths will be removed once the AWS AuthNZ is released to all environments.  So this is a workaround to prevent having to phase in the WSM fix to use user-facing ID for the `vwbusr:terra-workspace-id` tag.

### DIFF
--- a/startupscript/aws/vm-metadata.sh
+++ b/startupscript/aws/vm-metadata.sh
@@ -2,27 +2,37 @@
 
 # vm-metadata.sh
 #
-# Defines a function to retrieve an instance tag set on the VM.
+# Defines functions to retrieve an instance tag set on the VM.
 #
 # Note that this script is intended to be sourced from the "post-startup.sh" script
 # and is dependent on some functions and packages already installed.
 #
 # - aws (cli from ghcr.io/devcontainers/features/aws-cli:1) 
 
-function get_metadata_value() {
-  if [[ -z "$1" ]]; then
-    echo "usage: get_metadata_value <tag>"
+function get_metadata_value_raw() {
+    if [[ -z "$1" ]]; then
+    echo "usage: get_metadata_value_raw <tag>"
     exit 1
   fi
-  local tag_key=vwbusr:"$1"
 
   local imds_token
   imds_token="$(wget --method=PUT --header "X-aws-ec2-metadata-token-ttl-seconds:600" -q -O - http://169.254.169.254/latest/api/token)"
   local instance_id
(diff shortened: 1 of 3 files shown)


startupscript/aws/vm-metadata.sh
```diff
@@ -2,27 +2,37 @@
...
 # Note that this script is intended to be sourced from the "post-startup.sh" script
 # and is dependent on some functions and packages already installed.
 #
 # - aws (cli from ghcr.io/devcontainers/features/aws-cli:1) 
 
-function get_metadata_value() {
-  if [[ -z "$1" ]]; then
-    echo "usage: get_metadata_value <tag>"
+function get_metadata_value_raw() {
+    if [[ -z "$1" ]]; then
```
mbookman, line 13: This (new) indentation doesn't look right.

startupscript/aws/vm-metadata.sh
```diff
@@ -2,27 +2,37 @@
...
   local instance_id
   instance_id="$(wget --header "X-aws-ec2-metadata-token: ${imds_token}" -q -O - http://169.254.169.254/latest/meta-data/instance-id)"
   aws ec2 describe-tags \
-    --filters "Name=resource-id,Values=${instance_id}" "Name=key,Values=${tag_key}" \
+    --filters "Name=resource-id,Values=${instance_id}" "Name=key,Values=$1" \
     --query "Tags[0].Value" --output text 2>/dev/null
 }
+readonly -f get_metadata_value_raw
+
+function get_metadata_value() {
```
mbookman, line 28: Can you add comments (for each function or in the script header) explaining why there are two functions and when one would use the `_raw` version? There's something here about keeping nomenclature similar to the GCP version, but I'm not completely clear.

I'm always a little wary of things named `_raw` that maybe they should have a different name. In this case, should this be `get_metadata_tag_value` or some such?

mbookman (changes requested):

startupscript/aws/vm-metadata.sh
jmczerk: Added a comment block explaining that this is temporary code, and what it is working around.  PTAL.

yuhuyoyo (approved):

mbookman (approved):
LGTM

PR #99, opened 2024-05-20
Enable non-interactive login path for AWS.

jmczerk:
Enables non-interactive configuration in AWS on par with GCP when no-cred AWS authentication is enabled in the environment.

Key changes:
- Breaks out `aws-vault` config into separate script for reuse.
- Skip all of the extra config steps required when AWS login is enabled (we will be able to remove a bunch of code once AWS no-cred authn has propagated to all envs).
- Use direct call to git instead of `wb git clone --all`.  The CLI "passthrough" commands have a fundamental assumption that Google App Default Creds are configured, which only holds in GCP VM's.  This is pretty baked into the CLI's "external command" design, even though `git` doesn't use ADC, and would require some significant refactoring in CLI to change.  Scripting the same behavior is cloud-agnostic and trivial, so did that here.

### DIFF
--- a/startupscript/aws/resource-mount.sh
+++ b/startupscript/aws/resource-mount.sh
@@ -9,6 +9,7 @@
 # and is dependent on some functions and variables already being set up and some packages already installed:
 #
 # - emit (function)
+# - CLOUD_SCRIPT_DIR: path where AWS-specific scripts live
 # - LOG_IN: whether the user is logged into workbench CLI
 # - RUN_AS_LOGIN_USER: run command as user
 
@@ -25,5 +26,9 @@ else
 fi
 
 if [[ "${LOG_IN}" == "true" ]]; then
-  ${RUN_AS_LOGIN_USER} "wb resource mount"
+  source "${CLOUD_SCRIPT_DIR}/configure-aws-vault.sh"
+  ${RUN_AS_LOGIN_USER} "export AWS_VAULT_BACKEND='file' && \
+    export AWS_VAULT_FILE_PASSPHRASE='' && \
+    eval  \$(wb workspace configure-aws) && \
+    wb resource mount"
 fi
--- a/startupscript/aws/resource-mount.sh
+++ b/startupscript/aws/resource-mount.sh
@@ -27,5 +27,8 @@ fi
 
 if [[ "${LOG_IN}" == "true" ]]; then
   source "${CLOUD_SCRIPT_DIR}/configure-aws-vault.sh"
-  ${RUN_AS_LOGIN_USER} "source ${HOME}/.bashrc && wb resource mount"
+  ${RUN_AS_LOGIN_USER} "export AWS_VAULT_BACKEND='file' && \
+    export AWS_VAULT_FILE_PASSPHRASE='' && \
+    eval  \$(wb workspace configure-aws) && \
+    wb resource mount"
 fi
--- a/startupscript/aws/resource-mount.sh
+++ b/startupscript/aws/resource-mount.sh
@@ -27,8 +27,5 @@ fi
 
 if [[ "${LOG_IN}" == "true" ]]; then
   source "${CLOUD_SCRIPT_DIR}/configure-aws-vault.sh"
-  ${RUN_AS_LOGIN_USER} "export AWS_VAULT_BACKEND='file' && \
-    export AWS_VAULT_FILE_PASSPHRASE='' && \
-    eval  \$(wb workspace configure-aws) && \
-    wb resource mount"
+  ${RUN_AS_LOGIN_USER} "source ${HOME}/.bashrc && wb resource mount"
 fi
--- a/startupscript/git-setup.sh
+++ b/startupscript/git-setup.sh
@@ -50,7 +50,9 @@ ${RUN_AS_LOGIN_USER} "mkdir -p '${WORKBENCH_GIT_REPOS_DIR}'"
 # This loop replaces the logic of "wb git clone --all", which currently does not work without
 # Google Application Default Credentials being available.  This emulates the behavior of the CLI
 # command, continuing with an error message when an individual repo cannot be cloned.
-pushd "${WORKBENCH_GIT_REPOS_DIR}" || exit
+
+# shellcheck disable=SC2164
+pushd "${WORKBENCH_GIT_REPOS_DIR}"
 ${RUN_AS_LOGIN_USER} "wb resource list --type=GIT_REPO --format json" | \
   jq -c .[] | \
   while read -r ITEM; do
@@ -61,4 +63,5 @@ ${RUN_AS_LOGIN_USER} "wb resource list --type=GIT_REPO --format json" | \
         echo "git clone of ${GIT_REPO_URL} failed."
     fi
   done
-popd || exit
+# shellcheck disable=SC2164
+popd
--- a/startupscript/git-setup.sh
+++ b/startupscript/git-setup.sh
@@ -46,5 +46,14 @@ ${RUN_AS_LOGIN_USER} "mkdir -p '${WORKBENCH_GIT_REPOS_DIR}'"
 # to the git references, the corresponding git repo cloning will be skipped.
 # Keep this as last thing in script. There will be integration test for git cloning (PF-1660). If this is last thing, then
 # integration test will ensure that everything in script worked.
-${RUN_AS_LOGIN_USER} "cd '${WORKBENCH_GIT_REPOS_DIR}' && wb git clone --all"
-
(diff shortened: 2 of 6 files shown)


startupscript/setup-bashrc.sh
```diff
@@ -72,4 +72,23 @@ EOF
 
 fi
 
+if [[ "${CLOUD}" == "aws" && "${LOG_IN}" == "true" ]]; then
+
+  # Create a symlink to this workspace's AWS config file to use as the target for AWS_CONFIG_FILE.
+  readonly AWS_CONFIG_SYMLINK="${USER_WORKBENCH_CONFIG_DIR}/workspace.conf"
```
yuhuyoyo, line 78: can you add the new env var USER_WORKBENCH_CONFIG_DIR to the documentation after line 20

startupscript/git-setup.sh
```diff
@@ -46,5 +46,14 @@ ${RUN_AS_LOGIN_USER} "mkdir -p '${WORKBENCH_GIT_REPOS_DIR}'"
 # to the git references, the corresponding git repo cloning will be skipped.
 # Keep this as last thing in script. There will be integration test for git cloning (PF-1660). If this is last thing, then
 # integration test will ensure that everything in script worked.
-${RUN_AS_LOGIN_USER} "cd '${WORKBENCH_GIT_REPOS_DIR}' && wb git clone --all"
-
+pushd "${WORKBENCH_GIT_REPOS_DIR}" || exit 1
+${RUN_AS_LOGIN_USER} "wb resource list --type=GIT_REPO --format json" | \
+  jq -c .[] | \
```
yuhuyoyo, line 51: oh is git clone --all broken

startupscript/aws/resource-mount.sh
```diff
@@ -25,5 +26,9 @@ else
 fi
 
 if [[ "${LOG_IN}" == "true" ]]; then
-  ${RUN_AS_LOGIN_USER} "wb resource mount"
+  source "${CLOUD_SCRIPT_DIR}/configure-aws-vault.sh"
+  ${RUN_AS_LOGIN_USER} "export AWS_VAULT_BACKEND='file' && \
```
yuhuyoyo, line 30: source ~/.bashrc?

yuhuyoyo (approved):

startupscript/git-setup.sh
jmczerk: I added some notes about this in the PR description.  It's not really "broken"; looking at the CLI code it's working-as-intended, as the CLI treats git as a "passthrough" command and all passthrough commands require that Application Default Credentials are available.  Git doesn't actually make use of ADC, but we would have to special-case it not to behave this way.

So we could change the CLI behavior, but it seems like a design change rather than a bugfix, so I would rather not gate AWS support on this.  I think there's probably scope for re-visiting the whole "passthrough command" design in the CLI, but it's not trivial.

I'd also rather not branch the script when this logic works for both clouds.  WDYT?

yuhuyoyo: i see.  ya that sounds good to me to do it separately

so all the generic toolings are broken for aws instance?

jmczerk: Yeah, though other than `git`, I'm not sure why a VM user (AWS or GCP) would run any of them.  They could app-default login with `gcloud` on AWS to get access to GCP resources, without having to wrap them in `wb` calls.

I think the passthrough apps are mainly for devs to run tools as pet SA's on their laptops?  That's why I'm really hesitant to mess with the app behavior without more investigation; I don't understand if/how users are using them for commands other than git.

startupscript/aws/resource-mount.sh
jmczerk: `source` doesn't work inside `sudo`, rolled this back.

startupscript/git-setup.sh
```diff
@@ -46,5 +46,14 @@ ${RUN_AS_LOGIN_USER} "mkdir -p '${WORKBENCH_GIT_REPOS_DIR}'"
 # to the git references, the corresponding git repo cloning will be skipped.
 # Keep this as last thing in script. There will be integration test for git cloning (PF-1660). If this is last thing, then
 # integration test will ensure that everything in script worked.
-${RUN_AS_LOGIN_USER} "cd '${WORKBENCH_GIT_REPOS_DIR}' && wb git clone --all"
-
+pushd "${WORKBENCH_GIT_REPOS_DIR}" || exit 1
```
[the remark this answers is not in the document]
jmczerk: Without this lint triggers: https://www.shellcheck.net/wiki/SC2164

Should I use a directive to ignore this given we're running with `set -o errexit`?  Either way it should probably be `|| exit` (i.e. drop the `1`).  Preference?

startupscript/git-setup.sh
```diff
@@ -46,5 +46,14 @@ ${RUN_AS_LOGIN_USER} "mkdir -p '${WORKBENCH_GIT_REPOS_DIR}'"
...
+${RUN_AS_LOGIN_USER} "wb resource list --type=GIT_REPO --format json" | \
+  jq -c .[] | \
+  while read -r ITEM; do
+    GIT_REPO_NAME="$(echo "$ITEM" | jq -r .id)"
+    GIT_REPO_URL="$(echo "$ITEM" | jq -r .gitRepoUrl)"
+    if [[ ! -d "${GIT_REPO_NAME}" ]]; then
```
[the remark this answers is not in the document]
jmczerk: My read of the CLI code is that it will continue if one clone fails: https://github.com/verily-src/terra-tool-cli/blob/main/src/main/java/bio/terra/cli/command/app/passthrough/Git.java#L100

I will make this loop follow the same convention and add documentation.

startupscript/git-setup.sh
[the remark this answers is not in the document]
mbookman: That shellcheck _feels_ unnecessary to me for our startup scripts. These scripts run on machines that are in well-known states. It just doesn't seem (to me) to come up often that our startup scripts `cd` to directories that don't exist. My preference would be to ignore that one.

startupscript/git-setup.sh
jmczerk: Created [BENCH-3026](https://verily.atlassian.net/browse/BENCH-3026).

startupscript/git-setup.sh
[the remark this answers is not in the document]
jmczerk: Tested that this WAI (continues when a single clone fails).

startupscript/git-setup.sh
```diff
@@ -46,5 +46,22 @@ ${RUN_AS_LOGIN_USER} "mkdir -p '${WORKBENCH_GIT_REPOS_DIR}'"
...
 
+# This loop replaces the logic of "wb git clone --all", which currently does not work without
+# Google Application Default Credentials being available.  This emulates the behavior of the CLI
+# command, continuing with an error message when an individual repo cannot be cloned.
+
+# shellcheck disable=SC2164
+pushd "${WORKBENCH_GIT_REPOS_DIR}"
+${RUN_AS_LOGIN_USER} "wb resource list --type=GIT_REPO --format json" | \
+  jq -c .[] | \
+  while read -r ITEM; do
```
mbookman, line 58: There's really (IMO) nasty feature of Bash whereby:

```
<cmd> | while read ...; do
           <cmds>
        done
```

creates a subshell for the cmds to run in. Bash does not exit the out script with failure in the subshell.
You observed this in that a "git clone" failures do not exit this `while ... do` nor the script as a whole.

https://lists.gnu.org/archive/html/bug-bash/2008-01/msg00121.html

There are better patterns to follow as described here in the Google shell style guide:

https://lists.gnu.org/archive/html/bug-bash/2008-01/msg00121.html

We should *consider* enforcing the use of these other patterns to avoid the unexpected error behavior.

mbookman (approved):
LGTM. One comment added for future considerations.

startupscript/git-setup.sh
jmczerk: Just to clarify my previous comment, I did observe the undesired behavior (exit script on failed clone) before I added the `|| echo "git clone of ${GIT_REPO_URL} failed.""` to the `git clone` line inside the loop to allow it to continue.

mbookman: Interesting! I'll see if I can surface more info later, but I'm glad to see that (and I just did some testing that allays my fears here).

The issue of variable changes not propagating outward due to the subshell is real, and we should be aware of that, but this doesn't appear to be an immediate concern.

mbookman (approved):
LGTM

PR #100, opened 2024-05-21
Fix remounting on restart to work in both clouds.

jmczerk:
One more change that escaped my initial testing.  I tested this with custom apps for each of the touched devcontainer configs and it remounts successfully.

### DIFF
--- a/startupscript/remount-on-restart.sh
+++ b/startupscript/remount-on-restart.sh
@@ -40,9 +40,9 @@ readonly POST_STARTUP_OUTPUT_FILE="${USER_WORKBENCH_CONFIG_DIR}/remount-on-resta
 exec >> "${POST_STARTUP_OUTPUT_FILE}"
 exec 2>&1
 
-#######################################
-# Emit a message with a timestamp
-#######################################
+##############################
+# Import utility functions
+##############################
 source "${SCRIPT_DIR}/emit.sh"
 
 #############################
--- a/startupscript/remount-on-restart.sh
+++ b/startupscript/remount-on-restart.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# remount-on-restart.sh
+#
+# Remounts buckets for the logged in user when a devcontainer instance is restarted.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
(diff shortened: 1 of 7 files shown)


startupscript/remount-on-restart.sh
```diff
@@ -0,0 +1,53 @@
...
+# remount-on-restart.sh
+#
+# Remounts buckets for the logged in user when a devcontainer instance is restarted.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
+if [[ $# -ne 4 ]]; then
```
mbookman, line 12: Would be preferable for the 4th arg to be more of an enum like "login/nologin" and the downstream variable be more descriptive of what is being logged into (cloud, workbench, github).
I realize that's a more intrusive change than just fixing what you are currently working on, but would be a good sweep to make at some point.

startupscript/remount-on-restart.sh
```diff
@@ -0,0 +1,53 @@
...
+##################################################################
+# Send stdout and stderr from this script to a file for debugging.
+##################################################################
+readonly USER_WORKBENCH_CONFIG_DIR="${WORK_DIRECTORY}/.workbench"
+readonly POST_STARTUP_OUTPUT_FILE="${USER_WORKBENCH_CONFIG_DIR}/remount-on-restart-output.txt"
+exec >> "${POST_STARTUP_OUTPUT_FILE}"
+exec 2>&1
+
+#######################################
+# Emit a message with a timestamp
```
mbookman, line 44: The comment implies that the script emits a message here, rather than bringing in a utility function.
I'd be more inclined to just have this section comment be something like:

```
##############################
# Import utility functions
##############################
```

mbookman (approved):
A few suggestions, but nothing I feel strongly about.

LGTM

startupscript/remount-on-restart.sh
jmczerk: Agreed!  [BENCH-2566](https://verily.atlassian.net/browse/BENCH-2566) captures removing all the `login`-conditional code, at which point we'll remove the arg entirely.

mbookman (approved):
LGTM

PR #101, opened 2024-05-23
get the latest between the boot time and the last active time and set it to the metadata 

yuhuyoyo:
only use boottime when it's the latest of the three timestamp.

### DIFF
--- a/startupscript/butane/idle-shutdown.sh
+++ b/startupscript/butane/idle-shutdown.sh
@@ -28,12 +28,8 @@ fi
 
 # Check uptime first because on VM reboot, the last active timestamp could still be really old. Only start checking for last active timestamp
 # when the uptime is longer than the idle timeout threshold. 
-UP_TIME="$(awk '{print $1}' /proc/uptime)"
+UP_TIME="$(awk '{print int($1)}' /proc/uptime)"
 readonly UP_TIME
-if echo "${UP_TIME}" "${IDLE_TIMEOUT_SECONDS}" | awk '{if ($1 < $2) exit 0; else exit 1}'; then
-    emit "The VM has not been up for long enough to shut down. Uptime seconds: ${UP_TIME}. Idle timeout threshold is: ${IDLE_TIMEOUT_SECONDS}"
-    exit 0
-fi
 
 # Get the last time the VM was active. Default to 0 if not set.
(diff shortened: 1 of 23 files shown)


startupscript/butane/idle-shutdown.sh
```diff
@@ -46,6 +39,11 @@ LAST_ACTIVE_PROXY="$(get_guest_attribute "last-active/proxy" "0")"
 if [[ "${LAST_ACTIVE}" -lt "${LAST_ACTIVE_PROXY}" ]]; then
     LAST_ACTIVE="${LAST_ACTIVE_PROXY}"
 fi
+# get the latest time between the last boot time and the last active time. The last active timestamp could still be really old when the VM is rebooted and
+# the last active timestamp is not updated yet.
+if [[ "${LAST_ACTIVE}" -lt "${UP_TIME}" ]]; then
```
mbookman, line 44: The [Google Style Guide](https://google.github.io/styleguide/shellguide.html) doesn't specify a preference:

> Use (( … )) or -lt and -gt for numerical comparison.

but readers might find this more familiar syntactically:

```
if (( LAST_ACTIVE < UP_TIME )); then
```

startupscript/butane/idle-shutdown.sh
```diff
@@ -46,6 +39,11 @@ LAST_ACTIVE_PROXY="$(get_guest_attribute "last-active/proxy" "0")"
 if [[ "${LAST_ACTIVE}" -lt "${LAST_ACTIVE_PROXY}" ]]; then
     LAST_ACTIVE="${LAST_ACTIVE_PROXY}"
 fi
+# get the latest time between the last boot time and the last active time. The last active timestamp could still be really old when the VM is rebooted and
+# the last active timestamp is not updated yet.
```
jmczerk, line 43: Since `UP_TIME` is number of seconds since boot and `LAST_ACTIVE` is a timestamp, shouldn't we calculate based on `BOOT_TIME="$((NOW - UP_TIME))"` instead of `UP_TIME` on these two lines?

yuhuyoyo: right.. sorry made the change too fast

jmczerk (approved):
LGTM

PR #104, opened 2024-06-20
add a script for cleaning up metadata

yuhuyoyo:
this needs to go in first.
then i change verily1 to use it. 
then i update the deployment to use this in the downstream app-manger repo.

### DIFF
--- a/startupscript/butane/probe-user-access.sh
+++ b/startupscript/butane/probe-user-access.sh
@@ -29,7 +29,7 @@ if [[ "$(docker container inspect -f '{{.State.Running}}' "${CONTAINER_NAME}")" 
     fi
 fi
 
-LOAD="$(awk '{print $1}' /proc/loadavg)" # 1-minute average load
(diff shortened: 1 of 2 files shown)


startupscript/butane/probe-user-access.sh
```diff
@@ -29,7 +29,7 @@ if [[ "$(docker container inspect -f '{{.State.Running}}' "${CONTAINER_NAME}")"
     fi
 fi
 
-LOAD="$(awk '{print $1}' /proc/loadavg)" # 1-minute average load
```
yuhuyoyo, line 32: 1 min seems to be too noisy.

cbookg: 3 == 15 mins?

cbookg (approved):

startupscript/butane/probe-user-access.sh
yuhuyoyo: ya there're three options. 1 min, 5 min and 15 min

PR #117, opened 2024-08-14
Enable CLI installation on libreoffice

cbookg:
Changes include (mainly for alpine Linux):
- Using APK package manager instead of APT.
- Installing gcsfuse from source as it is not available for APK.

### DIFF
--- a/startupscript/gcp/resource-mount.sh
+++ b/startupscript/gcp/resource-mount.sh
@@ -4,6 +4,9 @@
 #
 # Installs gcsfuse if it is not already installed and mount resources.
 #
+# Both apt and install-from-source steps are based on gcloud docs here:
+# https://cloud.google.com/storage/docs/gcsfuse-install
+#
 # Note that this script is intended to be sourced from the "post-startup.sh" script
 # and is dependent on some functions and variables already being set:
 #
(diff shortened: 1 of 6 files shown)


startupscript/gcp/resource-mount.sh
```diff
@@ -11,21 +11,35 @@
...
 
-  # install packages needed to install gcsfuse
-  apt-get install -y \
-    fuse \
-    lsb-release
+    apk add --no-cache fuse git go
 
-  # Install based on gcloud docs here https://cloud.google.com/storage/docs/gcsfuse-install.
-  GCSFUSE_REPO="gcsfuse-$(lsb_release -c -s)"
-  readonly GCSFUSE_REPO
```
mbookman, line 23: Can you add a reference to where these install steps come from? Is there a doc like was referenced in the existing comment.

mbookman (changes requested):
LGTM. One small question.

startupscript/gcp/resource-mount.sh
cbookg: Oh its from the same page, but the comment is moved below into the "apt" section.  I can put it at the top or soemthing.

cbookg: updated comment

mbookman (approved):
LGTM

mbookman (approved):
LGTM

PR #129, opened 2024-10-11
BENCH-4272: Retry post-startup script commands that may fail due to to transient network issues

june-hua:


startupscript/post-startup.sh
```diff
@@ -14,13 +14,15 @@ readonly USER_NAME="${1}"
 readonly WORK_DIRECTORY="${2}"
```
june-hua, line 14: Exporting the variables that are used by the scripts that are called later. This is because those scripts cannot be sourced if we support retries.
Exporting the variables will make the variables available to those scripts.

startupscript/install-cli.sh
```diff
@@ -16,6 +16,11 @@
 # - USER_BASH_COMPLETION_DIR: path to the bash completion file
 # - LOG_IN: whether to log in to CLI
 
+set -o errexit
+set -o nounset
+set -o pipefail
+set -o xtrace
+
```
yuhuyoyo, line 23: check if cli is installed and only install if it's not

startupscript/post-startup.sh
```diff
@@ -29,10 +31,12 @@ source "${SCRIPT_DIR}/emit.sh"
 source "${CLOUD_SCRIPT_DIR}/vm-metadata.sh"
 
 readonly RUN_AS_LOGIN_USER="sudo -u ${USER_NAME} bash -l -c"
+export RUN_AS_LOGIN_USER
 
 # Startup script status is propagated out to VM guest attributes
 readonly STATUS_ATTRIBUTE="startup_script/status"
 readonly MESSAGE_ATTRIBUTE="startup_script/message"
+EXPORT STATUS_ATTRIBUTE MESSAGE_ATTRIBUTE
```
yuhuyoyo, line 39: what is export?

june-hua: Whoopsies let me fix that!

startupscript/post-startup.sh
yuhuyoyo: i see... can we do it per line.

like readonly USER_NAME
export USER_NAME

so when someone wants to delete or change the variable name, it's harder to miss.

yuhuyoyo (approved):

startupscript/post-startup.sh
```diff
@@ -155,13 +199,13 @@ source "${SCRIPT_DIR}/bash-completion.sh"
...
 if [[ "${LOG_IN}" == "true" ]]; then
-    source "${SCRIPT_DIR}/git-setup.sh"
+    retry 5 "${SCRIPT_DIR}/git-setup.sh"
 fi
 
 #############################
 # Mount buckets
 #############################
-source "${CLOUD_SCRIPT_DIR}/resource-mount.sh"
+retry 5 "${CLOUD_SCRIPT_DIR}/resource-mount.sh"
```
rogerwangcs, line 208: Don't think we should retry resource-mount. Non-zero exit codes are expected, so we'd treat those as success.

rogerwangcs (approved):

PR #131, opened 2024-10-21
Yyu fuse

yonghaoy:


src/jupyter-aou/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - NET_ADMIN
+  gcsfuse:
+    container_name: "gcsfuse"
+    image: "google/cloud-sdk:latest"
+    restart: always
+    volumes:
+      - gcsfuse-mount:/mnt/gcs
+    entrypoint: >
+      /bin/sh -c "
+      gcsfuse YOUR_BUCKET_NAME /mnt/gcs
```
yonghaoy, line 37: todo - pass YOUR_BUCKET_NAME as env var from Mikey vm init script

src/jupyter-aou/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
+version: '2.4'
+services:
+  app:
+    container_name: "application-server"
+    image: "gcr.io/deeplearning-platform-release/tf-gpu.2-11.py310"
```
yonghaoy, line 5: deep learning image is the AoU is using

src/jupyter-aou/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+    command: "start-notebook.py --ServerApp.token='' --ServerApp.password='' --ServerApp.root_dir=/home/jovyan/work --ServerApp.allow_origin='*'"
+    networks:
+      - app-network
+    devices:
+      - /dev/fuse
+    security_opt:
+      - apparmor:unconfined
+  wondershaper:
+    container_name: "wondershaper"
+    image: "us.gcr.io/broad-dsp-gcr-public/wondershaper:0.0.1"
```
yonghaoy, line 24: Wondershaper for rate limiting. Eventually, it's better to host this image in a VWB's public repo

src/jupyter-aou/docker-compose.yaml
```diff
@@ -0,0 +1,50 @@
...
+    restart: always
+    network_mode: "host"
+    cap_add:
+      - NET_ADMIN
```
yonghaoy, line 28: Net_admin is required by wondershaper sidecar

src/jupyter-aou/docker-compose.yaml
rogerwangcs: Is pytorch not available for AoU VMs?

rogerwangcs: Where/how are `terra-docker` images used? https://github.com/DataBiosphere/terra-docker

rogerwangcs: nvm, I see that this image is used as the base for all of their images https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/Dockerfile#L3

src/jupyter-aou/docker-compose.yaml
rogerwangcs: This is a cool exploration. I think the workbench way is to use the workbench CLI and run `wb resource mount`, which mounts all resource buckets in the workspace. This means we may need to build another image that specifically does this. The other problem is that now that users cannot remount buckets themselves, we'll need a cron in the mount sidecar that tries to remount every hour or so.

PR #134, opened 2024-10-30
BENCH-4471: Add jupyter template app

rogerwangcs:
This jupyter template app further parametrizes the container port and image in `docker-compose.yaml`. Add a few more sed lines to `parse-devcontainer.sh` to set the port and image fields provided by the generated butane script. See dependent app manager change: https://github.com/verily-src/verily1/pull/40902


features/src/jupyter/install.sh
```diff
@@ -0,0 +1,981 @@
...
+OVERRIDE_DEFAULT_VERSION="${OVERRIDEDEFAULTVERSION:-"true"}"
+
+export PIPX_HOME=${PIPX_HOME:-"/usr/local/py-utils"}
+
+### BEGIN: Workbench-specific customizations ###
+# Set the user to the provided feature option `USER`
+USERNAME="${USERNAME:-"${USER:-"automatic"}"}"
+# Set the cloud platform to the provided feature option `CLOUDPLATFORM`
+CLOUD_PLATFORM="${CLOUDPLATFORM:-"gcp"}"
+### END: Workbench-specific customizations ###
```
rogerwangcs, line 32: one change here

features/src/jupyter/install.sh
```diff
@@ -0,0 +1,981 @@
...
+        add_user_jupyter_config $CONFIG_DIR $CONFIG_FILE "c.NotebookApp.notebook_dir = '/home/${USERNAME}'"
+    fi
+
+    # If we are in a GCE environment, install the BigQuery Jupyter plugin, which
+    # includes jupyter lab UI plugins for viewing BQ tables, GCS buckets, and
+    # dataproc clusters.
+    if [ "${CLOUD_PLATFORM}" = "gcp" ]; then
+        install_user_package "${INSTALL_UNDER_ROOT}" "bigquery-jupyter-plugin"
+    fi
+### END: Workbench-specific customizations ###
```
rogerwangcs, line 975: The other change

features/src/jupyter/install.sh
```diff
@@ -0,0 +1,991 @@
...
+        add_user_jupyter_config "$CONFIG_DIR" "$CONFIG_FILE" "c.NotebookApp.notebook_dir = '/home/${USERNAME}'"
+    fi
+
+    # If we are in a GCE environment, install the BigQuery Jupyter plugin, which
+    # includes jupyter lab UI plugins for viewing BQ tables, GCS buckets, and
+    # dataproc clusters.
+    if [ "${CLOUD_PLATFORM}" = "gcp" ]; then
+        install_user_package "${INSTALL_UNDER_ROOT}" "bigquery-jupyter-plugin"
+    fi
+### END: Workbench-specific customizations ###
```
rogerwangcs, line 985: the other change

src/jupyter-template/devcontainer-template.json
```diff
@@ -0,0 +1,33 @@
...
+      "description": "Whether to log in to workbench CLI",
+      "proposals": ["true", "false"],
+      "default": "false"
+    },
+    "containerImage": {
+      "type": "string",
+      "description": "The container image to use",
+      "default": "debian:bullseye"
+    },
+    "containerPort": {
```
yuhuyoyo, line 26: is this changing depending on the image?

yuhuyoyo (approved):

features/src/jupyter/install.sh
```diff
@@ -0,0 +1,991 @@
+#!/usr/bin/env bash
+#-------------------------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
yuhuyoyo, line 3: removethe copyright

features/src/jupyter/install.sh
```diff
@@ -0,0 +1,991 @@
...
+# Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
+#-------------------------------------------------------------------------------------------------------------
+#
+# Docs: https://github.com/microsoft/vscode-dev-containers/blob/main/script-library/docs/python.md
+# Maintainer: The VS Code and Codespaces Teams
```
yuhuyoyo, line 8: remove maintainer

features/src/jupyter/install.sh
rogerwangcs: Is that usually how it works when we make modifications? Idk how code licensing works

src/jupyter-template/devcontainer-template.json
yuhuyoyo: dont' quote me but this is what we did for app proxy stuff

rogerwangcs: Yeah, we currently have a port field in app attributes which we use to set the port for the proxy agent https://github.com/verily-src/verily1/pull/40902/files#diff-88d46f9da504f046fab382aae3a9feb1f9c5ee6676faff4fafff281eef543aa3R48, this change is just to pipe that into the container port too.

yuhuyoyo: sorry i meant to respond to the other thread about copyright...

rogerwangcs:
FYI @yuhuyoyo , I made a change to the smoke test script that fixed it to replace every devcontainer template option, also enabled jupyter-template in tests. The code you have for substituting options in the test is nice, why don't we just use it in `parse-devcontainer.sh` 😆

rogerwangcs:
> sure. now the templating is getting more real and complicated.

You mentioned before how we're in a weird state where we kind of follow devcontainer template patterns so it maybe could be published in the future. But I think we will need more complex file subsitutions in the future, like adding the config fields for GPU support in the docker compose or the new application level logging stuff that @june-hua proposed. So we might break devcontainer template compatibility and just use our own scripting here

yuhuyoyo:
the devcontainer cli templating support setting options in all the files in the .devcontainer directory iirc

rogerwangcs:
> the devcontainer cli templating support setting options in all the files in the .devcontainer directory iirc

yeah that makes sense, I saw an example in their repo that had a template option in a docker compose. But in general template options only good for simple var value subsitutions. Docker compose also doesn't handle conditional blocks well. Technically it's possible, but it would be messy within our setup. I think I will need to conditionally set the `deploy: resources: reservations: device: gpu blah` block to make GPU work.

yuhuyoyo:
> > the devcontainer cli templating support setting options in all the files in the .devcontainer directory iirc
> 
> yeah that makes sense, I saw an example in their repo that had a template option in a docker compose. But in general template options only good for simple var value subsitutions. Docker compose also doesn't handle conditional blocks well. Technically it's possible, but it would be messy within our setup. I think I will need to conditionally set the `deploy: resources: reservations: device: gpu blah` block to make GPU work.

i don't know what you mean and how this is relevant to parse-devcontainer.sh

rogerwangcs:
> i don't know what you mean and how this is relevant to parse-devcontainer.sh

Yeah not atm, but will be soon 😄

PR #137, opened 2024-11-05
Add prebuilt image devcontainer template

rogerwangcs:
Adds a skeleton devcontainer that does not install any additional features with the expectation that the container image URI provided is contains all the needed workbench dependencies such as aws-cli, gcloud cli, java, jupyterlab.

Followup work: Parametrize the login user so that this devcontainer user can be used for any workbench prebuilt app, not just jupyter lab.


src/prebuilt-jupyter-template/.devcontainer.json
```diff
@@ -0,0 +1,54 @@
+{
+  "name": "Prebuilt Workbench JupyterLab devcontainer template",
```
yuhuyoyo, line 2: just a naming suggestion. 

custom-workbench-jupyter-template

i think custom is a more popular term to describe them.

yuhuyoyo (approved):

PR #145, opened 2024-11-27
BENCH-4629: Conditionally insert nvidia runtime block in docker-compose

rogerwangcs:
This change modifies the `parse-devcontainer.sh` script to accept an accelerator param to conditionally insert a gpu runtime block in the docker compose file of the devcontainer. This block tells docker to reserve nvidia GPU devices in the VM.

The script is rewritten to be idempotent so that on reruns (in the future), we can determine whether the gpu reservation block should be set or not set depending if the VM has a gpu attached. This to get us one step closer to making gpu editability work with Mikey apps.

### DIFF
--- a/startupscript/butane/005-parse-devcontainer.sh
+++ b/startupscript/butane/005-parse-devcontainer.sh
@@ -13,6 +13,7 @@ function usage {
   echo "  devcontainer_path: folder directory of the devcontainer."
   echo "  cloud: gcp or aws."
   echo "  login: whether the user is logged into the workbench on startup."
+  echo "  accelerator: the accelerator to use. E.g: nvidia"
   echo "  container_image: the container image to use."
   echo "  container_port: the port to expose."
   exit 1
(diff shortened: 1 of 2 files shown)


startupscript/butane/nvidia-runtime.yaml
```diff
@@ -0,0 +1,11 @@
+# Name: nvidia-runtime.yaml
+# Description: This snippet is used to reserve all GPUs on the host machine for the NVIDIA runtime.
+# Start of snippet from startupscript/butane/nvidia-runtime.yaml
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              count: all
+              capabilities: [gpu]
```
rogerwangcs, line 10: After trying a bunch of different things, I couldn't make a static docker compose work, which is really unfortunate. But at least this block here is basically all we need to tell docker that the container needs access to nvidia gpu devices on the host, so in a way, this is better than having to paste some GPU block into every devcontainer that wants to be able to use gpus.

startupscript/butane/005-parse-devcontainer.sh
```diff
@@ -46,15 +68,42 @@ replace_template_options() {
...
+
+echo 'publishing devcontainer.json to metadata'
 export PATH="/opt/bin:$PATH"
 # shellcheck source=/dev/null
-source /home/core/metadata-utils.sh
+source '/home/core/metadata-utils.sh'
 readonly JSONC_STRIP_COMMENTS=/home/core/jsoncStripComments.mjs
 DEVCONTAINER_CUSTOMIZATIONS="$("${JSONC_STRIP_COMMENTS}" < "${DEVCONTAINER_CONFIG_PATH}" | jq -c .customizations.workbench)"
 readonly DEVCONTAINER_CUSTOMIZATIONS
-set_metadata "devcontainer/customizations" "${DEVCONTAINER_CUSTOMIZATIONS}"
```
cbookg, line 60: why single quotes?

startupscript/butane/005-parse-devcontainer.sh
```diff
@@ -46,15 +68,42 @@ replace_template_options() {
...
 replace_template_options "${DEVCONTAINER_CONFIG_PATH}"
 replace_template_options "${DEVCONTAINER_DOCKER_COMPOSE_PATH}"
 
-echo "publishing devcontainer.json to metadata"
+# apply gpu runtime block if accelerator is nvidia
+if [[ "${ACCELERATOR}" == "nvidia" ]]; then
+    apply_gpu_runtime "${DEVCONTAINER_DOCKER_COMPOSE_PATH}" "${NVIDIA_RUNTIME_PATH}"
+fi
+
+echo 'publishing devcontainer.json to metadata'
```
cbookg, line 102: why single quotes?

startupscript/butane/nvidia-runtime.yaml
rogerwangcs: docker compose up failure, I tried various values here too 😔

startupscript/butane/005-parse-devcontainer.sh
```diff
@@ -26,14 +27,35 @@ fi
...
+readonly NVIDIA_RUNTIME_PATH="${DEVCONTAINER_PATH}/startupscript/butane/nvidia-runtime.yaml"
 
 readonly DEVCONTAINER_CONFIG_PATH="${DEVCONTAINER_PATH}/.devcontainer.json"
 readonly DEVCONTAINER_DOCKER_COMPOSE_PATH="${DEVCONTAINER_PATH}/docker-compose.yaml"
 
-if [[ -d /home/core/devcontainer/startupscript ]]; then
-    cp -r /home/core/devcontainer/startupscript "${DEVCONTAINER_PATH}"/startupscript
+# On first run, copy existing .devcontainer.json and docker-compose.yaml to template files
+# On subsequent runs, the template files will be used to replace the original files
+# so that if arguments change, they are properly applied to the original template files
```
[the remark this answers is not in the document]
rogerwangcs: June is working on GPU editability where we want users to be able to stop, attach or detach gpus to their VMs. We'll probably want the accelerator argument to be dynamic in the future (maybe read from vm metadata) to enable this.

cbookg (approved):

PR #148, opened 2024-12-17
BENCH-4825: Set instance region gcloud config on post startup

rogerwangcs:
The Mikey custom app has a BQ/GCS Jupyterlab plugin that depends on this config being set.

### DIFF
--- a/startupscript/gcp/vm-metadata.sh
+++ b/startupscript/gcp/vm-metadata.sh
@@ -2,8 +2,17 @@
 
 # vm-metadata.sh
 
-# Defines a function to retrieve an instance attributes set on the VM.
+# Retrieve the instance zone.
+function get_instance_region() {
+  curl --retry 5 -s -f \
+    -H "Metadata-Flavor: Google" \
+    "http://metadata.google.internal/computeMetadata/v1/instance/zone" | \
+    awk -F'/' '{print $4}' | \
+    sed 's/-[^-]*$//'
+}
+readonly -f get_instance_region
 
+# Retrieves an instance attributes on the VM.
 function get_metadata_value() {
   if [[ -z "$1" ]]; then
     echo "usage: get_metadata_value <tag>"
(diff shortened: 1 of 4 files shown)


startupscript/gcp/vm-metadata.sh
```diff
@@ -2,8 +2,17 @@
 
 # vm-metadata.sh
 
-# Defines a function to retrieve an instance attributes set on the VM.
+# Retrieve the instance zone.
+function get_instance_region() {
```
yuhuyoyo, line 6: can you modify the description that this is retrieve the instance zone and convert it to a region by removing the zonal suffix.

rogerwangcs: Done!

startupscript/post-startup.sh
```diff
@@ -74,8 +74,8 @@ cd /tmp || exit
 # Make the .workbench directory as the user so that they own it and have correct linux permissions.
 ${RUN_AS_LOGIN_USER} "mkdir -p '${USER_WORKBENCH_CONFIG_DIR}'"
 ${RUN_AS_LOGIN_USER} "ln -sf '${USER_WORKBENCH_CONFIG_DIR}' '${USER_WORKBENCH_LEGACY_CONFIG_DIR}'"
-exec >> "${POST_STARTUP_OUTPUT_FILE}"
-exec 2>&1
+exec > >(tee -a "${POST_STARTUP_OUTPUT_FILE}")  # Append output to the file and print to terminal
+exec 2> >(tee -a "${POST_STARTUP_OUTPUT_FILE}" >&2)  # Append errors to the file and print to terminal
```
rogerwangcs, line 78: I think we should split stdout/stderr to echo as well, I'm tired of looking at post-startup-output.txt every single time. This should give us better logs both for the smoke tests here, and when looking at the journalctl logs when iterating on a devcontainer.

yuhuyoyo: agree. sgtm

yuhuyoyo (approved):

PR #159, opened 2025-03-10
fix app proxy for non-google

yuhuyoyo:


rogerwangcs (commented):
Remind me again, does this work for all of the other dataproc components? Hive/hadoop/flink etc?

yuhuyoyo:
> Remind me again, does this work for all of the other dataproc components? Hive/hadoop/flink etc?

i'm just fixing the url path, it has no work related..

but i handed this off to Andrew who is doing some proxy stuff.

PR #164, opened 2025-03-25
AoU Jupyter App

pantherman594:
Adds a devcontainer template for the AoU Jupyter app

BENCH-5437

### DIFF
--- a/src/jupyter-aou/docker-compose.yaml
+++ b/src/jupyter-aou/docker-compose.yaml
@@ -1,0 +1,31 @@
+version: "2.4"
+services:
+  app:
+    container_name: "application-server"
+    image: "${templateOption:containerImage}"
+    user: "jupyter"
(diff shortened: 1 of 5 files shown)


src/jupyter-aou/docker-compose.yaml
```diff
@@ -0,0 +1,31 @@
...
+    user: "jupyter"
+    restart: always
+    volumes:
+      - .:/workspace:cached
+    ports:
+      - "8888:8888"
+    networks:
+      - app-network
+    cap_add:
+      - SYS_ADMIN
```
rogerwangcs, line 15: So sounds like we decided the existing resource mount setup is OK security wise?

rogerwangcs: Sounds good.

PR #149, opened 2025-01-08
[DO NOT MERGE] Testing a jupyter app with docker daemon mounted

rogerwangcs:
https://verily.atlassian.net/browse/PHP-45008

PR #171, opened 2025-04-17
create workbench-jupyter-docker devcontainer to support docker workflows

npc-g:
Supports Docker out of Docker (DooD) by exposing the docker socket and running jupyter under the docker group.

Mostly #149 with some small changes from #163

### DIFF
--- a/src/workbench-jupyter-docker/.devcontainer.json
+++ b/src/workbench-jupyter-docker/.devcontainer.json
@@ -4,6 +4,8 @@
   "service": "app",
   "shutdownAction": "none",
   "workspaceFolder": "/workspace",
+  // Get the host's docker group ID and propagate it into the .env file, which
+  // allows it to be used within docker-compose.yaml.
   "initializeCommand": "DOCKER_GID=`getent group docker | cut -d: -f3` && echo \"DOCKER_GID=${DOCKER_GID}\" > .env",
   "postCreateCommand": "./startupscript/post-startup.sh jupyter /home/jupyter ${templateOption:cloud} ${templateOption:login} && ./startupscript/setup-docker.sh",
   // re-mount bucket files on container start up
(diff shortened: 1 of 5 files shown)


src/workbench-jupyter-docker/.devcontainer.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "Workbench JupyterLab with docker support devcontainer template",
+  "dockerComposeFile": "docker-compose.yaml",
+  "service": "app",
+  "shutdownAction": "none",
+  "workspaceFolder": "/workspace",
+  "initializeCommand": "DOCKER_GID=`getent group docker | cut -d: -f3` && echo \"DOCKER_GID=${DOCKER_GID}\" > .env",
```
rogerwangcs, line 7: nit: comment explaining this init command?

startupscript/setup-docker.sh
```diff
@@ -0,0 +1,45 @@
...
+
+###########################
+# Configuring docker auth #
+###########################
+
+# Give user write permissions to the mounted docker config directory
+chown -R jupyter /home/jupyter/.docker
+
+# Login to docker with gcloud credentials (needs to be re-run every 30 min if needed)
+sudo -u jupyter /bin/bash -c "docker login -u oauth2accesstoken -p $(gcloud auth print-access-token) https://us-central1-docker.pkg.dev"
```
rogerwangcs, line 45: I can't remember why I chose the access token route. Gcp supports a number of other auth methods. If rerunning every 30 min is not desirable, we could experiment with gcloud credential helper https://cloud.google.com/artifact-registry/docs/docker/authentication

rogerwangcs (approved):

startupscript/setup-docker.sh
npc-g: IIRC we tried `gcloud auth configure-docker` but it was making changes to the docker config file within the container rather than in the host, so went with this approach instead

PR #172, opened 2025-04-18
remove sudo pw requirement for vscode user abc

yuhuyoyo:
BENCH-5530


startupscript/post-startup.sh
```diff
@@ -202,6 +202,25 @@ if [[ "${LOG_IN}" == "true" ]]; then
...
+
+# Check if there's an old rule in the main sudoers file that requires a password
+if grep -q "^${USER_NAME} ALL=(ALL:ALL) ALL" "${sudoers_file}"; then
+  echo "Found password-requiring rule for ${USER_NAME} in /etc/sudoers. Commenting it out."
+  
+  # Comment out the old rule in /etc/sudoers
+  sed -i "s/^${USER_NAME} ALL=(ALL:ALL) ALL/# ${USER_NAME} ALL=(ALL:ALL) ALL/" "${sudoers_file}"
+fi
+
+echo "${USER_NAME} ALL=(ALL) NOPASSWD:ALL" > "${sudoers_d_file}"
```
pantherman594, line 219: IMO this should be done conditionally (maybe another parameter for this startup script command? or maybe more generally we can have this startup script check for/run all scripts in a custom startup scripts directory which the docker compose for vscode can mount)

There are apps like the AoU jupyter app that we may not want to give users sudo access

startupscript/post-startup.sh
```diff
@@ -202,6 +202,25 @@ if [[ "${LOG_IN}" == "true" ]]; then
...
+chmod 440 /etc/sudoers.d/"${USER_NAME}"
```
june-hua, line 220: nit: use ${sudoers_d_file}

startupscript/post-startup.sh
yuhuyoyo: i like having a custom directory

june-hua:
Should these changes also be extended to JupyterLab?

src/vscode/.devcontainer.json
```diff
@@ -4,13 +4,8 @@
...
   "workspaceFolder": "/workspace",
-  "postCreateCommand": [
-    "./startupscript/post-startup.sh",
-    "abc",
-    "/config",
-    "${templateOption:cloud}",
-    "${templateOption:login}"
-  ],
+  "postCreateCommand":
+    "./startupscript/post-startup.sh abc /config ${templateOption:cloud} ${templateOption:login}; ./sudo-passwordless.sh abc",
```
pantherman594, line 8: nit: it might be better for the `templateOption`s to be double quoted so that they can't get expanded into multiple values

pantherman594 (approved):

pantherman594:
> Should these changes also be extended to JupyterLab?

The JupyterLab docker image should already be set up to give sudo without needing a password so I don't think it'll be necessary there. This vscode image we're using uniquely has a password requirement for sudo for some reason

yuhuyoyo:
> > Should these changes also be extended to JupyterLab?
> 
> The JupyterLab docker image should already be set up to give sudo without needing a password so I don't think it'll be necessary there. This vscode image we're using uniquely has a password requirement for sudo for some reason

ya, same with r-analysis. already support sudo.

pantherman594 (approved):

PR #176, opened 2025-04-29
BENCH-6277: Login user on VM startup before resource remount

june-hua:


startupscript/remount-on-restart.sh
```diff
@@ -45,6 +45,13 @@ exec 2>&1
 ##############################
 source "${SCRIPT_DIR}/emit.sh"
 
+#############################
+# CLI login
+#############################
+if [[ "${LOG_IN}" == "true" ]] && wb auth status 2>&1 | grep -q "NO USER LOGGED IN"; then
```
yuhuyoyo, line 51: we should start cleaning up LOG_IN. i believe it should be true everywhere now.

yuhuyoyo (approved):

startupscript/remount-on-restart.sh
june-hua: Will remove

rogerwangcs: What was the LOG_IN flag for initially? I always wondered about it

june-hua: Hm it's set to false in the CI test - which I think is why the tests are failing. Do you know why it's set to false there?

PR #178, opened 2025-05-12
Move fuse commands to a sidecar for jupyter-aou

pantherman594:
This forwards gcsfuse, goofys, and fusermount commands via ssh to the remotefuse sidecar, which operates on a mount shared with the main container. With this, the main container does not need CAP_SYS_ADMIN

PHP-70786

### DIFF
--- a/src/jupyter-aou/docker-compose.yaml
+++ b/src/jupyter-aou/docker-compose.yaml
@@ -44,9 +44,6 @@ services:
       - SYS_ADMIN
     devices:
       - /dev/fuse
-    environment:
-      CLOUD: ${templateOption:cloud}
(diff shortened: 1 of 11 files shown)


src/jupyter-aou/docker-compose.yaml
```diff
@@ -10,25 +10,45 @@ services:
     restart: always
     volumes:
       - .:/workspace:cached
+      - /tmp/wb-mount:/home/jupyter/workspace:slave
```
yuhuyoyo, line 13: dumb question, where is /tmp/wb-mount?

yuhuyoyo (approved):

PR #185, opened 2025-05-29
Add custom extensions for AoU

pantherman594:
Adds a custom AoU extension to provide a policy reminder prompt when downloading or uploading files. This accomplishes the same thing as https://github.com/verily-src/verily1/pull/63993 but in the devcontainer.

Most of `extension/aou-jupyterlab` was generated using the [JupyterLab extension template](https://github.com/jupyterlab/extension-template/)

This is based off of https://github.com/all-of-us/workbench/blob/main/api/src/main/webapp/static/aou-file-tree-policy-extension.js, but had to be rewritten since the JupyterLab extension system is completely different from Jupyter Notebook's.

This also adds a snippets extension with a sample snippet

image

image

image

PHP-62591


src/jupyter-aou/build/snippets/All of Us Cromwell Setup Python snippets/(1) Setup.py
```diff
@@ -0,0 +1,130 @@
+import os
+import subprocess
+import json
+import requests
+from pathlib import Path
+
+def check_for_app(env):
+    list_apps_url = f'{env["leonardo_url"]}/api/google/v1/apps/{env["google_project"]}'
```
yuhuyoyo, line 8: how are the snippet used? this is not valid for us

src/jupyter-aou/build/snippets/All of Us Cromwell Setup Python snippets/(1) Setup.py
```diff
@@ -0,0 +1,130 @@
+import os
```
yuhuyoyo, line 1: does the snippets folder has to be space separated?

src/jupyter-aou/build/snippets/All of Us Cromwell Setup Python snippets/(1) Setup.py
pantherman594: I just imported all of the ones they had before, I'm not sure which snippets they'll want to keep or change but I'm not worried about that for now.

I could also just add a sample snippet and import later as needed, would you prefer that?

src/jupyter-aou/build/snippets/All of Us Cromwell Setup Python snippets/(1) Setup.py
pantherman594: no, but the name of the folder corresponds to the submenu name

image

yuhuyoyo: oh i see.

PR #186, opened 2025-05-30
PHP-67559: Install Cromwell, Nextflow and dsub on VM at first startup

june-hua:


startupscript/post-startup.sh
```diff
@@ -202,6 +204,11 @@ if [[ "${LOG_IN}" == "true" ]]; then
     retry 5 "${SCRIPT_DIR}/git-setup.sh"
 fi
 
+###################################
+# Install Workflow tools
+###################################
+source "${SCRIPT_DIR}/install-workflow-tools.sh"
```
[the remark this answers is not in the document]
pantherman594: Oh 15 seconds shouldn't be a problem, thanks for checking

pantherman594 (approved):

june-hua:
> > @yuhuyoyo @pantherman594 Do either of you know where we define what libs are installed for these CI tests? Looks like `pip` is not available; neither is `pythom -m venv`
> > I also see`python venv` is [not available by default in vscode](https://verily.atlassian.net/browse/PHP-63139), which means I will need to add that in if I decide to go the virtual env route.
> 
> CI test is just running the devcontainer with auth off. just install python as a devcontainer feature with venv? https://github.com/devcontainers/features/tree/main/src/python

Thanks! I see now that python is a hard dependency for debian (which is why the vs-code test passes) but is a soft dependency for ubuntu (which is why the other tests fail).
I'll add python as a devcontainer feature.

However, Jupyter and R-analysis have python pre-installed on VMs despite their corresponding `.devcontainer.json` not specifying Debian, therefore falling back on Ubuntu. Do you know why this is?

yuhuyoyo:
> > > @yuhuyoyo @pantherman594 Do either of you know where we define what libs are installed for these CI tests? Looks like `pip` is not available; neither is `pythom -m venv`
> > > I also see`python venv` is [not available by default in vscode](https://verily.atlassian.net/browse/PHP-63139), which means I will need to add that in if I decide to go the virtual env route.
> > 
> > 
> > CI test is just running the devcontainer with auth off. just install python as a devcontainer feature with venv? https://github.com/devcontainers/features/tree/main/src/python
> 
> Thanks! I see now that python is a hard dependency for debian (which is why the vs-code test passes) but is a soft dependency for ubuntu (which is why the other tests fail). I'll add python as a devcontainer feature.
> 
> However, Jupyter and R-analysis have python pre-installed on VMs despite their corresponding `.devcontainer.json` not specifying Debian, therefore falling back on Ubuntu. Do you know why this is?

what do you mean falling back on Ubuntu?

the app image for jupyter probably already have python

startupscript/install-workflow-tools.sh
```diff
@@ -0,0 +1,60 @@
...
+
+emit "Installing Nextflow ..."
+retry 5 install_nextflow
+mv nextflow "${USER_HOME_LOCAL_BIN}"
+
+#######################################
+# Install dsub
+#######################################
+function install_dsub() {
+  ${RUN_AS_LOGIN_USER} "pip install --user dsub"
```
yuhuyoyo, line 55: should we use venv for pip install?

startupscript/install-workflow-tools.sh
```diff
@@ -0,0 +1,73 @@
...
+
+readonly VENV_PATH="${WORK_DIRECTORY}/.venv"
+readonly DSUB_VENV_PATH="${VENV_PATH}/dsub_libs"
+mkdir -p VENV_PATH
+
+function install_dsub() {
+  ${RUN_AS_LOGIN_USER} "${DSUB_VENV_PATH}/bin/pip install dsub"
+}
+
+apt install -y python3-venv
```
june-hua, line 64: Need to install python3-venv

startupscript/install-workflow-tools.sh
```diff
@@ -0,0 +1,73 @@
...
+PYTHON_COMMAND=$(command -v python3)
```
june-hua, line 65: At this point, the root PATH has the python3 path. But the user's PATH does not.

startupscript/install-workflow-tools.sh
```diff
@@ -0,0 +1,73 @@
...
+${RUN_AS_LOGIN_USER} "${PYTHON_COMMAND} -m venv ${DSUB_VENV_PATH}"
+retry 5 install_dsub
+
+# Add convenience variable & alias
+cat << EOF >> "${USER_BASHRC}"
+export DSUB_VENV_PATH='${DSUB_VENV_PATH}'
+alias dsub_activate='source ${DSUB_VENV_PATH}/bin/activate'
```
yuhuyoyo, line 72: do you mind adding to this test to check for installation of these tools? https://github.com/verily-src/workbench-app-devcontainers/blob/master/test/test.sh#L8

june-hua: I needed to add the env variables to a new file ~/.bash_env for the test script. Need the variables to check if cromwell and dsub are installed.

Couldn't source ~/.bashrc in the test script because bashrc exits early for non-interactive shell.

PR #198, opened 2025-06-27
fix the script for devcontainer parser

yuhuyoyo:
1. devcontainer.json can either lives in .devcontainer/devcontainer.json or just .devcontainerjson in the root project. Currently it only handles the latter.

2. there may not be a docker-compose.yaml file.

PHP-74142


rogerwangcs (approved):
Do we have devcontainers that aren't using docker compose? I'm curious if GPU setup needs to be configured differently for those cases since we're been doing it via docker compose.

yuhuyoyo:
> Do we have devcontainers that aren't using docker compose? I'm curious if GPU setup needs to be configured differently for those cases since we're been doing it via docker compose.

we don't but customer can have

PR #216, opened 2025-08-06
Add scripts to enable running of wb on Flatcar host.

jmczerk:
Similar to the recommended usage of `aws`/`gcloud` in Flatcar, we would also like to be able to run our CLI via `docker run`, instead of having to manually install the dependencies, which will add complexity and hurt startup time).


startupscript/butane/wb/configure.sh
```diff
@@ -0,0 +1,51 @@
...
+# 2. Authenticating with the Workbench server
+# 3. Setting up the default workspace context
+#
+# The script uses metadata values to configure:
+# - terra-cli-server: The Workbench server endpoint
+# - terra-workspace-id: The default workspace to use
+#
+# This is typically run once during environment setup to preconfigure
+# the workbench for subsequent interactive use.
+
```
yuhuyoyo, line 17: can this be part of the 00x- script so it's clear when it's executed in the app start up systemd.

yuhuyoyo (approved):

startupscript/butane/wb/configure.sh
yuhuyoyo: ah sad will need to update [this ](https://github.com/verily-src/verily1/blob/90f0e11b0de2880ee83683720d3abb155692a446/workbench/app-manager/static/butane/butane.yaml.tmpl#L97). just fyi we bump the version automatically with a pr once a day and i'm ok dev is broken for a short bit.

PR #220, opened 2025-08-13
Pin remotefuse to Debian 12

pantherman594:
gcsfuse isn't available for Debian 13 yet

### DIFF
--- a/src/jupyter-aou/remotefuse/Dockerfile
+++ b/src/jupyter-aou/remotefuse/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:latest
+FROM debian:12-slim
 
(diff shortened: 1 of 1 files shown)


pantherman594:
I ran the app image CD after merging the cache project IAM changes and it failed to install gcsfuse

yuhuyoyo:
> I ran the app image CD after merging the cache project IAM changes and it failed to install gcsfuse

we should add a test here for building the image

pantherman594:
The CI test does build the image, it's included in the jupyter-aou devcontainer. This only became an issue because debian:latest changed from 12 to 13 yesterday at 4 pm. Do you think we should run the smoke test nightly as well?

PR #230, opened 2025-09-03
Enable installFromSource for workbench-tools feature in GCP apps

pantherman594:
Adds a new cloud parameter to specify the cloud the app is running on. If the cloud is "gcp", installFromSource will default to true for workbench-tools.

PHP-71105


june-hua (approved):
Why is this not installed from source by default for AWS?

PR #231, opened 2025-09-04
Set WORKBENCH_INSTALL_PATH when remounting after container restart

pantherman594:
remount-on-restart assumes WORKBENCH_INSTALL_PATH is set which is not true when the container is restarting after a prior create

### DIFF
--- a/startupscript/remount-on-restart.sh
+++ b/startupscript/remount-on-restart.sh
@@ -14,6 +14,9 @@ if [[ $# -ne 4 ]]; then
   exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


startupscript/remount-on-restart.sh
```diff
@@ -14,6 +14,9 @@ if [[ $# -ne 4 ]]; then
   exit 1
 fi
 
+readonly WORKBENCH_INSTALL_PATH="${WORKBENCH_INSTALL_PATH:-/usr/bin/wb}"
+export WORKBENCH_INSTALL_PATH
```
yuhuyoyo, line 18: why export?

yuhuyoyo (approved):

PR #232, opened 2025-09-05
Always enable websocket transport

pantherman594:
PHP-90214


pantherman594:
nemo test failure is a known issue not from this PR

PR #237, opened 2025-09-17
Create AoU common build directory

pantherman594:


yuhuyoyo:
looks like nemo jupyter aou test failure is real

pantherman594:
> looks like nemo jupyter aou test failure is real

It's failing on `apt-get update` with error code 137 which usually means out of memory... still trying to come up with how to address that but open to suggestions

yuhuyoyo:
> > looks like nemo jupyter aou test failure is real
> 
> It's failing on `apt-get update` with error code 137 which usually means out of memory... still trying to come up with how to address that but open to suggestions

oh makes sense. nemo is already huge ..

PR #238, opened 2025-09-18
Delete wb container after it exits

pantherman594:
The `wb` script currently doesn't delete the container after it exits, leading to something like this:

image


yuhuyoyo (approved):
good catch

PR #367, opened 2026-04-07
Added back clis to RStudio

michaelxiao7:
- Reverting changes in https://github.com/verily-src/workbench-app-devcontainers/pull/357
- App default size was incremented in a separate PR to support these CLIs and prevent some OOM issues

PHP-135615


june-hua (approved):

pantherman594:
note that since the ghcr-cache change was reverted we might run into github rate limits again :(

PR #369, opened 2026-04-09
Add register-key script to register an app's public key

pantherman594:
Also renames startup scripts to add more space in between for inserting additional scripts

PHP-127691


startupscript/butane/035-register-key.sh
```diff
@@ -0,0 +1,140 @@
...
+
+# Get an access token via the Workbench CLI
+TOKEN="$(/home/core/wb.sh auth print-access-token)"
+readonly TOKEN
+
+echo "Checking signing algorithm for resource ${RESOURCE_ID}..."
+
+SIGNING_ALGORITHM="$(curl -s -f \
+  -H "Authorization: Bearer ${TOKEN}" \
+  "${WSM_URL}/api/workspaces/v1/${WORKSPACE_ID}/resources/controlled/gcp/gce-instances/${RESOURCE_ID}" \
```
michaelxiao7: Is this supposed to be hardcoded to `gcp/gce-instances` if AWS is also supported on line 52?

pantherman594: Whoops yes, good catch!

michaelxiao7 (approved):
LGTM!

PR #381, opened 2026-04-23
Add secret producer/receiver and docker credential helper

pantherman594:
See linked verily1 PR for ordering of when the scripts will be called.

- `docker-auth-secrets.sh` registers docker-credential-workbench-secret to be used when pulling a package from one of the "dockerRepository" secrets. This must be called after git-clone-devcontainer since it needs to read the secret from the devcontainer directory
- `docker-credential-secrets.sh` is the script behind docker-credential-workbench-secret (the docker-credential scripts are just a thin wrapper providing the resource path)
- `provide-secrets.sh` fetches secrets configured in secrets.yml and passes them to the app. This is the "Secret Provider" in the design doc
- `secret-receiver` receives the secret values and directs them to environment variables or file descriptors for the app to read
- `vscode-secrets` is a sample vscode app that reads a secret named `example-secret` and stores it in environment variable `EXAMPLE_SECRET`

PHP-127691


startupscript/butane/055-provide-secrets.sh
```diff
@@ -0,0 +1,198 @@
...
+  elif [[ -n "${PATH_VAR}" ]]; then
+    SECRET_TYPE="pathVar"
+    SECRET_TARGET="${PATH_VAR}"
+  elif [[ -n "${VALUE_VAR}" ]]; then
+    SECRET_TYPE="valueVar"
+    SECRET_TARGET="${VALUE_VAR}"
+  fi
+
+  # Look up secret's workspace and resource IDs from attached secrets map
+  SECRET_WORKSPACE_ID="$(echo "${ATTACHED_SECRETS}" | jq -r --arg name "${SECRET_NAME}" '.[$name].workspaceId')"
```
june-hua, line 147: is this workspace id always the same as ${WORKSPACE_ID} from above?

startupscript/butane/055-provide-secrets.sh
```diff
@@ -0,0 +1,198 @@
...
+done
+
+echo "Delivering ${PIPE_SECRET_COUNT} secret(s) to container..."
+
+set +o xtrace
+if ! echo "${SECRETS_JSON}" | timeout 30 docker exec -i "${CONTAINER_NAME}" sh -c "cat > ${PIPE_PATH}"; then
+  >&2 echo "ERROR: Timed out writing secrets to container pipe."
+  exit 1
+fi
+
```
june-hua, line 186: `unset SECRETS_JSON`
but value will still be in memory.

could also stream them one by one to pipe so not all secrets will be in memory at the same time

startupscript/butane/055-provide-secrets.sh
```diff
@@ -0,0 +1,198 @@
...
+RESOURCE_ID="$(get_metadata_value "wb-resource-id" "")"
+readonly RESOURCE_ID
+if [[ -z "${RESOURCE_ID}" ]]; then
+  echo "No resource ID found in metadata. Skipping secret provisioning."
+  exit 0
+fi
+
+set +o xtrace
```
june-hua, line 85: nit: not necessary but may be nice to have a convenience wrapper like

sensitive_command TOKEN="$(/home/core/wb.sh auth print-access-token)"

startupscript/butane/secret-utils.sh
```diff
@@ -0,0 +1,126 @@
...
+  if [[ -z "${nonce}" || "${nonce}" == "null" ]]; then
+    >&2 echo "ERROR: Failed to get challenge nonce for secret ${secret_resource_id}."
+    return 1
+  fi
+
+  local signature
+  signature="$(sign_nonce "${nonce}" "${key_file}")"
+
+  local read_request
+  read_request="$(jq -n \
```
june-hua, line 64: `set +o xtrace` for all nonce related lines as well?

startupscript/butane/055-provide-secrets.sh
pantherman594: no, this is the workspace the secret resource belongs in, which will most likely be different from the workspace this app is in

startupscript/butane/secret-utils.sh
pantherman594: We could but they're not very sensitive information. Each (randomly generated) nonce is only valid for 30 seconds and can only be used once

startupscript/butane/055-provide-secrets.sh
pantherman594: unfortunately that won't help with xtrace :(

it'll evaluate wb auth print-access-token first, which gets printed, before it even gets to the wrapper

startupscript/butane/055-provide-secrets.sh
june-hua: Got it thanks for clarifying!

startupscript/butane/055-provide-secrets.sh
pantherman594: I don't know about streaming them one by one, that makes both ends more complicated (how do we signal that we're done and ready to start the main process?) and all secrets have to be in memory at the same time anyways since we don't want to persist them on disk

startupscript/butane/secret-utils.sh
```diff
@@ -0,0 +1,126 @@
...
+#   $2 - workspace ID of the attached secret
+#   $3 - resource ID of the attached secret
+# Returns: 0 if allowed (or no restriction), 1 if not allowed
+function validate_allowed_secret() {
+  local secret_entry="$1"
+  local workspace_id="$2"
+  local resource_id="$3"
+
+  local allowed_secrets
+  allowed_secrets="$(echo "${secret_entry}" | jq '.allowedSecrets // empty')"
```
june-hua, line 107: is this from secrets.json (which is converted from secrets.yml)?
The user will need to know the source workspace id and resource id?

startupscript/butane/055-provide-secrets.sh
june-hua: it could wait for some eof indicator to tell if streaming is done, but as you said the secrets would still be available in the container memory.. then it wouldn't improve security.

june-hua (approved):

startupscript/butane/secret-utils.sh
pantherman594: Yes its from secrets.json, no the user will not need to know the source workspace ID/resource ID, those will be set by the UI when creating an app with attached secrets

PR #386, opened 2026-04-29
Add SAS Analytics Pro devcontainer app for AoU

yonghaoy:
## Summary
- Adds the `aou-sas` devcontainer app running SAS Analytics Pro on VWB GCE
- Integrates with Mikey Secrets for secure SAS license delivery (`secrets.yml` + entrypoint wrapper), with bind-mount fallback for manual GCE testing
- Includes wondershaper sidecar (2MB/s bandwidth limit), Apache auto-login, CORS/CSRF/cookie config, gcsfuse, gcloud SDK, and apt shims for RHEL compatibility

## Files
| File | Purpose |
|------|---------|
| `.devcontainer.json` | Devcontainer spec with Workbench lifecycle hooks |
| `Dockerfile` | Minimal — stages load-envs, adds entrypoint wrapper |
| `docker-compose.yaml` | App service (ports 80/443) + wondershaper sidecar |
| `sas-entrypoint.sh` | Bridges Mikey Secrets fd → `/sasinside/SASLicense.jwt` |
| `sas-startup.sh` | PRE_DEPLOY_SCRIPT: packages, user setup, SAS config, Apache proxy |
| `secrets.yml` | Declares SAS license for Mikey Secrets |
| `sasinside/.gitignore` | Keeps dir in git, excludes license files |

## Test plan
- [ ] Deploy to GCE with Mikey Secrets enabled — verify license is delivered and SAS starts
- [ ] Deploy to GCE without Mikey Secrets — place license manually in `sasinside/`, verify fallback works
- [ ] Verify `aou` user cannot read `/sasinside/SASLicense.jwt`
- [ ] Verify wondershaper bandwidth limiting is active
- [ ] Verify SAS Studio auto-login (no login prompt)
- [ ] Verify gcsfuse bucket mounting works via Workbench startup scripts


src/aou-sas/.devcontainer.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "SAS App for AoU",
+  "dockerComposeFile": ["docker-compose.yaml", "../aou-common/aou-common-compose.yaml"],
+  "service": "app",
+  "runServices": ["app", "wondershaper"],
+  "shutdownAction": "none",
+  "workspaceFolder": "/workspace",
+  "initializeCommand": "umount -f /tmp/wb-mount/*; rm -rf /tmp/wb-mount/*",
```
pantherman594: this was a holdover from when I was trying to move gcsfuse to a sidecar, it's not needed anymore (and I should delete this from jupyter-aou as well...)

src/aou-sas/sas-startup.sh
```diff
@@ -0,0 +1,125 @@
...
+PROXY_CONF=/etc/httpd/conf.d/dkrapro-proxy.conf
+
+# Comment out default RequestHeader lines from the SAS image, then re-add
+# exactly the ones we need.  Use a marker so restarts are idempotent.
+if ! grep -q "AOU-CONFIGURED" "${PROXY_CONF}"; then
+  sed -i "s/RequestHeader/#RequestHeader/g" "${PROXY_CONF}"
+
+  # Force Apache to generate https:// URLs in redirects (RedirectMatch etc.)
+  # since the Workbench proxy terminates TLS upstream.
+  sed -i 's|^ServerName localhost$|ServerName https://localhost|' "${PROXY_CONF}"
```
pantherman594: I don't understand what this is doing, shouldn't we only be getting http requests by this point? since like the comment says the proxy terminates TLS upstream

src/aou-sas/sas-startup.sh
pantherman594, line 1: could most of these be part of the dockerfile? it seems to me like `Data directories` and `AoU environment loader` are the only ones that need to be done at runtime

src/aou-sas/docker-compose.yaml
```diff
@@ -0,0 +1,63 @@
...
+    build:
+      context: .
+      additional_contexts:
+        load-envs: service:aou-common-load-envs
+    restart: always
+    ports:
+      - "80:80"
+    volumes:
+      - .:/workspace:cached
+      - sas-data:/data
```
pantherman594, line 15: since the user's home directory is a volume, we should be careful not to persist the `$HOME/workspace` directory where resources will be mounted

this is how r-analysis does it
https://github.com/verily-src/workbench-app-devcontainers/blob/d1a9abc209902bc0349660a1ea5984233a4ab9e7/src/r-analysis/docker-compose.yaml#L9-L10

src/aou-sas/.devcontainer.json
```diff
@@ -0,0 +1,24 @@
...
+    "${templateOption:login}"
+  ],
+  "postStartCommand": [
+    "./startupscript/remount-on-restart.sh",
+    "aou",
+    "/data",
+    "${templateOption:cloud}",
+    "${templateOption:login}"
+  ],
+  "remoteUser": "root"
```
yonghaoy, line 22: should remove user be root?

pantherman594: nope this should be fine, this is the user the postcreate/poststart commands are run as, the docker container is still run as whichever user is configured in the image

src/aou-sas/sas-startup.sh
yonghaoy: There is actually an apachy proxy as side car of SAS. Apache uses it to generate https:// in redirect Location headers (not for handling incoming requests)

src/aou-sas/.devcontainer.json
yonghaoy: Done — removed `initializeCommand` from both `aou-sas` and `jupyter-aou` devcontainer.json files.

src/aou-sas/sas-startup.sh
yonghaoy: Good question — this doesn't affect how Apache handles incoming requests (which are indeed HTTP). It controls how Apache generates `Location` headers in redirect responses (`RedirectMatch`, `Redirect` directives). Without `ServerName https://`, Apache constructs `http://` redirect URLs. The Workbench proxy rewrites the host but not the scheme, so browsers would receive `http://` redirect URLs that fail behind the HTTPS proxy. Improved the comment to clarify this (now in the Dockerfile since the whole Apache config block moved there).

src/aou-sas/sas-startup.sh
yonghaoy: Done — moved package installs (apt shims, system packages, gcsfuse, gcloud SDK), user creation, SAS configuration, and Apache proxy config to the Dockerfile. `sas-startup.sh` now only handles runtime steps: data directory creation on the `/data` volume, license permissions, and copying the env loader to `/data`.

src/aou-sas/docker-compose.yaml
yonghaoy: Good catch — added `tmpfs: /data/workspace:uid=1001,gid=1001` following the r-analysis pattern. Also set explicit UID=1001 for the `aou` user in the Dockerfile to match.

src/aou-sas/sas-startup.sh
yonghaoy: My 2nd comment, replied by Claude haha

src/aou-sas/Dockerfile
```diff
@@ -6,6 +6,72 @@ USER root
 # This must be in the Dockerfile because it references a build context (load-envs).
 COPY --from=load-envs /dist/load-env /dist/load-env.sh /opt/sas/aou/
 
+###############################################################################
+# Package-manager compatibility
+# Workbench startup scripts (post-startup.sh, resource-mount.sh) expect
+# apt-get / apt.  These shims delegate to yum on this RHEL-based SAS image.
+###############################################################################
+RUN printf '#!/bin/bash\ncase "$1" in\n  update) exec yum makecache -y ;;\n  install) shift; exec yum install -y "$@" ;;\n  *) exec yum "$@" ;;\nesac\n' > /usr/local/bin/apt-get && \
```
pantherman594, line 14: we should probably update workbench startup scripts to handle either package manager, but for now if this works that's good!

src/aou-sas/secrets.yml
```diff
@@ -0,0 +1,3 @@
+secrets:
+  - name: "SASLicense"
```
pantherman594, line 2: I think we'll need another secret for docker credentials as well, but that could be a separate change after all my secrets stuff is in

pantherman594 (approved):
LGTM!

src/aou-sas/Dockerfile
yonghaoy: Agreed — a distro-agnostic package install helper in the shared startup scripts would be cleaner long-term. The shim works for now.

src/aou-sas/secrets.yml
yonghaoy: Sounds good — will add the docker credential secret in a follow-up once the secrets infrastructure is in place.

PR #387, opened 2026-04-29
Fluent-bit send application-server logs only

june-hua:
Exclude logs from other containers, e.g. app-proxy, fluent-bit


pantherman594:
I feel like this could still be useful to have stored in case we need to look through it in the future, what do you think about filtering out in the UI or Axon?

michaelxiao7 (approved):

PR #396, opened 2026-05-07
Add early access CDR config and GAR for AoU data collections

yonghaoy:
## Summary
- Adds `cdr_config_early_access.json` with registered tier (wb-crisp-citrus-2555, R2025Q4R4) and controlled tier (wb-bright-olive-4015, C2025Q4R3) entries
- Maps two new data collection UUIDs to the `early_access` cdrEnv in `datacollection_mapping.json`
- Wires up the new config in `configs.go` (embed) and `main.go` (switch case)

## Test plan
- [ ] Verify `go build ./...` passes in `src/aou-common/load-envs`
- [ ] Verify a workspace linked to the early access data collections resolves the correct CDR config
- [ ] Confirm env vars are correctly exported for both registered and controlled tiers

PHP-148755

### DIFF
--- a/src/aou-common/load-envs/configs/cdr_config_early_access.json
+++ b/src/aou-common/load-envs/configs/cdr_config_early_access.json
@@ -1,0 +1,31 @@
+{
+  "accessTiers": [
+    {
+      "shortName": "registered",
+      "datasetsBucket": "",
+      "artifactRegistryRepo": "us-central1-docker.pkg.dev/all-of-us-rw-prod/aou-rw-gar-remote-repo-docker-prod"
+    },
+    {
+      "shortName": "controlled",
+      "datasetsBucket": "gs://vwb-aou-earlyaccess-datasets-controlled/",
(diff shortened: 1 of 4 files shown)


yonghaoy:
@tarekmamdouh I would need your help to take pass those env vars (those are copied from prod ct so those might be wrong)? 

@pantherman594 I added our use app cache gar as env var, it should be okay?

pantherman594:
> I added our use app cache gar as env var, it should be okay?

That should be fine, but I don't think I see it in this PR?

pantherman594 (approved):
go changes LGTM

tarekmamdouh:
> @tarekmamdouh I would need your help to take pass those env vars (those are copied from prod ct so those might be wrong)?
> 
> @pantherman594 I added our use app cache gar as env var, it should be okay?

I update the config and pushed the change

tarekmamdouh (approved):

PR #410, opened 2026-05-13
Fix SAS startup issues, WB CLI, and environment variables

pantherman594:


src/aou-sas/sas-post-deploy.sh
```diff
@@ -0,0 +1,15 @@
...
+# be readable by the sas user during application (PRE_DEPLOY), but afterward
+# we restrict it to root so the aou user cannot exfiltrate it via pipe commands.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [ -f /sasinside/SASLicense.jwt ]; then
+  chmod 400 /sasinside/SASLicense.jwt
+  chown root:root /sasinside/SASLicense.jwt
```
june-hua, line 14: does the user have root access?

src/aou-sas/secrets.yml
```diff
@@ -2,4 +2,4 @@ secrets:
   - name: "SASDockerAuth"
     dockerRegistry: "cr.sas.com"
   - name: "SASLicense"
-    pathVar: "SAS_LICENSE_PATH"
+    pipeVar: "SAS_LICENSE_PATH"
```
june-hua, line 5: why was this switched to pipevar?

src/aou-sas/docker-compose.yaml
```diff
@@ -18,19 +18,17 @@ services:
...
+      - ./sas-post-deploy.sh:/opt/sas/aou/sas-post-deploy.sh:ro
     tmpfs:
       - /data/workspace:uid=1002,gid=1100
     environment:
       HOST_AUTH: ""
       SAS_DEBUG: "0"
       SASLICENSEFILE: "SASLicense.jwt"
       GOMEMLIMIT: "20MiB"
-      PRE_DEPLOY_SCRIPT: "bash /opt/sas/aou/sas-startup.sh"
+      PRE_DEPLOY_SCRIPT: "bash /opt/sas/aou/sas-pre-deploy.sh"
```
june-hua, line 30: are these used by the sas image?

pantherman594: yup, the SAS base image's startup script runs these

src/aou-sas/secrets.yml
pantherman594: sas-entrypoint.sh copies the license key to `/sasinside/SASLicense.jwt`, so we only need to read once. `pipeVar` and `pathVar` are pretty similar, except `pathVar` can be read multiple times which we don't need

src/aou-sas/sas-post-deploy.sh
pantherman594: nope! alternatively we could completely delete the SASLicense file since from my testing SAS only reads it on startup. I was worried it could unexpectedly delete the user's license file if `sasinside` is a mounted volume, but that might not be a huge concern

src/aou-sas/secrets.yml
june-hua: Gotcha, thanks for explaining!

june-hua (approved):

PR #411, opened 2026-05-13
Bench 8768 yp

PeterSu92:
## Summary

  Playground app has been 502ing because postgres 18 moved the default data directory from `/var/lib/postgresql/data` to a version-specific subdirectory (`/var/lib/postgresql/18/docker`). Our named volume was still mounted at the old path, so pg18 ignored it entirely, failed to init, and exited. Since the playground service depends on db being healthy, it never started either — Caddy couldn't resolve the upstream and returned 502s. Official docs: https://hub.docker.com/_/postgres

  Fix: mount the volume at `/var/lib/postgresql` instead, which is what the official postgres docker image now expects for 18+. This is also forward-compatible with future major
  versions.

  Ref: https://github.com/docker-library/postgres/pull/1259

  ## Test plan

  - [x] Spin up playground app, confirm db container starts and passes its healthcheck
  - [x] Confirm playground service comes up and is reachable through Caddy (no 502)

I tested both in a GCP workspace


pantherman594 (approved):

pantherman594:
Thank you!

PR #414, opened 2026-05-18
Run context gen script as jupyter user

june-hua:
The context gen script was failing because `wb` was not setup. The CLI is only setup for jupyter user, so run script as the user.

### DIFF
--- a/src/workbench-jupyter-with-llm/.devcontainer.json
+++ b/src/workbench-jupyter-with-llm/.devcontainer.json
@@ -13,7 +13,7 @@
     "${templateOption:login}"
   ],
   // re-mount bucket files on container start up, then generate LLM context
-  "postStartCommand": "./startupscript/remount-on-restart.sh jupyter /home/jupyter \"${templateOption:cloud}\" \"${templateOption:login}\"; /opt/llm-context/run-context-generator.sh /home/jupyter || true",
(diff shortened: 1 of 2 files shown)


june-hua:
Confirmed that fix works: 
image

michaelxiao7 (approved):

src/workbench-jupyter-with-llm/.devcontainer.json
```diff
@@ -13,7 +13,7 @@
     "${templateOption:login}"
   ],
   // re-mount bucket files on container start up, then generate LLM context
-  "postStartCommand": "./startupscript/remount-on-restart.sh jupyter /home/jupyter \"${templateOption:cloud}\" \"${templateOption:login}\"; /opt/llm-context/run-context-generator.sh /home/jupyter || true",
+  "postStartCommand": "./startupscript/remount-on-restart.sh jupyter /home/jupyter \"${templateOption:cloud}\" \"${templateOption:login}\"; sudo -u jupyter /opt/llm-context/run-context-generator.sh /home/jupyter || true",
```
NavidZ, line 16: No need to change the direction now. but more of a long term comment.
I think we should have added a dependency for that llm-feature on wb tools so devcontainer run/install that after wb tools so the wb is installed and also then in the ideal world have this generator run in the scripts within that feature and not here. So as an app author I would only need to include that feature in the "features" section and not additional commands here.

june-hua: For the `wb` CLI dependency, `workbench-tools` feature does not install it. It gets installed in `postCreateCommand` `post-startup.sh`.
We might consider moving that to `workbench-tools` in the future.

I also see that the context generation script is run on each devcontainer startup, curious if it only needs to be once on creation.

PR #418, opened 2026-05-21
SAS app health check retry

june-hua:
Retry for sas app health check up to 60 seconds. It may not be in healthy state yet right after devcontainer service completes

### DIFF
--- a/startupscript/butane/probe-proxy-readiness.sh
+++ b/startupscript/butane/probe-proxy-readiness.sh
@@ -11,43 +11,61 @@ set -o xtrace
 # shellcheck source=/dev/null
 source /home/core/metadata-utils.sh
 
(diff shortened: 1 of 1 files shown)


yonghaoy (approved):

yonghaoy:
A data point, I wait for few minutes until SAS get health after startup script...

june-hua:
> A data point, I wait for few minutes until SAS get health after startup script...

Hm, maybe I need to bump the retry count. I'll bump to 5 minutes

june-hua:
Tested out change, took ~1 minute for status to change from "UNHEALTHY" to "HEALTHY"

june-hua:
Ignore failures, they are due to image being too large, should be fixed separately

michaelxiao7 (approved):

PR #426, opened 2026-06-08
Implement lock for aws-vault for better parallelization

NBBense:
Add locking mechanism to prevent concurrent aws-vault processes.

BENCH-8769

### DIFF
--- a/startupscript/aws/configure-aws-vault.sh
+++ b/startupscript/aws/configure-aws-vault.sh
@@ -38,7 +38,7 @@ export AWS_VAULT_FILE_PASSPHRASE=""
 # do not use.
 export DBUS_SESSION_BUS_ADDRESS="/dev/null"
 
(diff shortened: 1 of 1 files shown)


startupscript/aws/configure-aws-vault.sh
```diff
@@ -38,6 +38,13 @@ export AWS_VAULT_FILE_PASSPHRASE=""
 # do not use.
 export DBUS_SESSION_BUS_ADDRESS="/dev/null"
 
+if [ "$1" = "export" ]; then
```
pantherman594:
```suggestion
if [ "\$1" = "export" ]; then
```

since we're using `cat <<EOF` to write to a file, `$1` needs to be escaped. Otherwise it'll be replaced with the first argument of configure-aws-vault

pantherman594 (approved):

PR #427, opened 2026-06-11
Install jupyter-server-proxy in AoU Jupyter image

yonghaoy:
## Summary
- Installs `jupyter-server-proxy` in the AoU Jupyter Docker image to enable `/proxy/<PORT>` URL routing
- This allows AoU Jupyter users to access dashboards and services (e.g. Streamlit, Dash, Shiny) running inside the container, matching the behavior of non-AoU Jupyter apps
- The existing `server-proxy-notif` extension already checks for this package at runtime and will automatically start working once it's available

## Context
The `/proxy/<PORT>` path works in non-AoU Jupyter apps because their base images include `jupyter-server-proxy`. The AoU base image (`app-aou-jupyter`) does not include it, so requests to `/proxy/<PORT>` fall through to the default JupyterLab handler and the user sees the JupyterLab UI instead of their dashboard.

**Open question:** Was `jupyter-server-proxy` intentionally omitted from AoU for security/compliance reasons, or was it simply not a requirement at the time? Please confirm before merging.

## Test plan
- [ ] Build the AoU Jupyter image and verify `jupyter-server-proxy` is installed (`pip show jupyter-server-proxy`)
- [ ] Start a service on an arbitrary port (e.g. `python -m http.server 8050`) and confirm `/proxy/8050/` routes correctly
- [ ] Verify the `server-proxy-notif` toast notification appears for newly detected ports

### DIFF
--- a/src/jupyter-aou/Dockerfile
+++ b/src/jupyter-aou/Dockerfile
@@ -1,5 +1,9 @@
 FROM us-west2-docker.pkg.dev/shared-pub-buckets-94mvrf/workbench-artifacts/app-aou-jupyter@sha256:54504cac626fa9ec61083b87e0c87f2d7859faf7a65830b2e8a743d6a08ac7a2
 
+# Enable /proxy/<PORT> routing so users can access dashboards and other
+# services running inside the container (e.g. Streamlit, Dash, Shiny).
+RUN pip install --no-cache-dir jupyter-server-proxy
+
 # Install jupyter extensions
 RUN --mount=type=bind,from=jupyter-extension-builder,source=/dist,target=/tmp/extensions \
     /tmp/extensions/setup.sh


michaelxiao7 (approved):

pantherman594:
Ideally this should be done in https://github.com/verily-src/verily1/blob/main/workbench/app-images/images/aou-jupyter/Dockerfile but I'd also like to investigate why it's not already included

yonghaoy:
## Investigation

Traced the issue through the image chain:

1. **`app-workbench-jupyter` (base image)** — `jupyter-server-proxy` **is** listed in `pixi.toml` and should be installed in the conda `jupyter` environment.

2. **`app-aou-jupyter`** ([`verily1/workbench/app-images/images/aou-jupyter/Dockerfile`](https://github.com/verily-src/verily1/blob/main/workbench/app-images/images/aou-jupyter/Dockerfile)) — This image is **pinned to an old base image tag**:
   ```dockerfile
   # Pin the workbench-jupyter image to a specific version for now, until
   # aou-jupyter's dependencies are also updated to use python 3.12.
   FROM $APP_REGISTRY_URL/app-workbench-jupyter:workbench-260323-1330-rc5 AS base-with-root
   ```
   It then runs a `mamba install` with a list of AoU-specific packages. This mamba solve may be **removing `jupyter-server-proxy`** as a side effect since it's not in the AoU package list, or the pinned base image tag predates when `jupyter-server-proxy` was added to `pixi.toml`.

### Recommended fix (in `verily1`)
Either:
- Add `jupyter-server-proxy` to the `mamba install` list in `aou-jupyter/Dockerfile`, or
- Unpin from the old base image (which already includes `jupyter-server-proxy`)

This PR in `workbench-app-devcontainers` works as a stopgap but the proper fix belongs upstream in the base image as @pantherman594 noted.

yonghaoy:
> Ideally this should be done in https://github.com/verily-src/verily1/blob/main/workbench/app-images/images/aou-jupyter/Dockerfile but I'd also like to investigate why it's not already included

https://github.com/verily-src/verily1/pull/126601 Looks it is installed in the base but then got override by AoU?

PR #431, opened 2026-06-16
Insert do not edit comment at the top of templated files

pantherman594:
These files should not be edited since they will be overwritten. We'll insert the note right after copying

### DIFF
--- a/startupscript/butane/050-parse-devcontainer.sh
+++ b/startupscript/butane/050-parse-devcontainer.sh
@@ -54,12 +54,14 @@ if [[ ! -f "${DEVCONTAINER_CONFIG_PATH}.template" ]]; then
 else
     cp "${DEVCONTAINER_CONFIG_PATH}.template" "${DEVCONTAINER_CONFIG_PATH}"
 fi
+sed -i "1s|^|// DO NOT EDIT: generated from ${DEVCONTAINER_CONFIG_PATH}.template on startup; edits will be overwritten.\n|" "${DEVCONTAINER_CONFIG_PATH}"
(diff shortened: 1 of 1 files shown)


startupscript/butane/050-parse-devcontainer.sh
```diff
@@ -54,12 +54,14 @@ if [[ ! -f "${DEVCONTAINER_CONFIG_PATH}.template" ]]; then
 else
     cp "${DEVCONTAINER_CONFIG_PATH}.template" "${DEVCONTAINER_CONFIG_PATH}"
 fi
+sed -i "1s|^|// DO NOT EDIT: generated from ${DEVCONTAINER_CONFIG_PATH}.template on startup; edits will be overwritten.\n|" "${DEVCONTAINER_CONFIG_PATH}"
```
june-hua, line 57: could we update the docker-compose files themselves?

NavidZ (commented):
Random comment that you don't need to worry about right now and maybe just keep somewhere in the backlog.

I think long term it might be better to start moving these parts of the scripts that are not necessary user modifiable (as oppose to the apps themselves or their startup scripts) to some internal repo (or at the very least separate from this repo).

startupscript/butane/050-parse-devcontainer.sh
pantherman594: do you mean adding this comment to files in this repo like `src/vscode/docker-compose.yaml`? Then the comments will stay in the .template files too

june-hua: Oh right ignore me

june-hua (approved):

PR #432, opened 2026-06-16
PHP-160287: Pin nextflow version

june-hua:
Nextflow v26 has incompatible config syntax. `wb nextflow` will be deprecated soon, so instead of supporting latest version, use the same nextflow version that workbench uses for workflows created in the UI.

PHP-160287

### DIFF
--- a/features/src/workbench-tools/install.sh
+++ b/features/src/workbench-tools/install.sh
@@ -88,7 +88,7 @@ readonly CONDA_PACKAGES_BINARIES=(
     "conda-forge::bgenix"
     "conda-forge::cromwell"
     "bioconda::ensembl-vep>=115"
-    "bioconda::nextflow=25.10.02"
+    "bioconda::nextflow=25.10.2"
     "bioconda::plink"
     "bioconda::plink2"
     "bioconda::regenie"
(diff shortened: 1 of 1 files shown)


JulesPetit (approved):

tshay54 (approved):

features/src/workbench-tools/install.sh
```diff
@@ -88,7 +88,7 @@ readonly CONDA_PACKAGES_BINARIES=(
     "conda-forge::bgenix"
     "conda-forge::cromwell"
     "bioconda::ensembl-vep>=115"
-    "bioconda::nextflow"
+    "bioconda::nextflow=25.10.02"
```
pantherman594: hmm I don't see `25.10.02` in the listed versions at https://anaconda.org/channels/bioconda/packages/nextflow/overview, should it be `25.10.2`?

pantherman594 (approved):

features/src/workbench-tools/install.sh
june-hua: .02 worked for me, it is equivalent to 25.10.2, but will update it so that it matches exactly

pantherman594: oh I guess it was still able to find/pick `25.10.2`

PR #435, opened 2026-06-17
fix(wb-mcp-server): Return detailed error messages instead of exit codes

anshisax:
Previously, when wb commands failed, the MCP server only returned generic exit codes (e.g., 'Error: exit status 1') without the actual error message.

Now includes command output with the error:

Testing with folder resource called "testingdelete" in verily internal testing workspace:
Before: Error: exit status 2
After:  Command failed with exit status 2:
        Error: invalid argument 'testingdelete' for '--id' flag: invalid UUID length: 14

### DIFF
--- a/features/src/wb-mcp-server/main.go
+++ b/features/src/wb-mcp-server/main.go
@@ -4,10 +4,8 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
-	"flag"
 	"fmt"
 	"io"
-	"log"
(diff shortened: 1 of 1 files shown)


pantherman594 (changes requested):
Hey Anshi, this is making a lot of changes beyond just returning detailed error messages. Could you check of those other changes are necessary, and either separate them out into another PR or update the title/description/scope of this one as needed?

PR #441, opened 2026-07-01
Configurable /dev/shm using metadata/guest attributes

pantherman594:
Uses metadata if available (for when we have it configurable from the frontend), falls back to guest attributes (which the user can set from inside the VM), then falls back to 64mb if neither are set

To set, curl from within the app:

```shell
curl -X PUT \
  -H "Metadata-Flavor: Google" \
  --data "256m" \
  "http://metadata.google.internal/computeMetadata/v1/instance/guest-attributes/config/shm-size"
```

PHP-161176

### DIFF
--- a/src/aou-sas/devcontainer-template.json
+++ b/src/aou-sas/devcontainer-template.json
@@ -15,6 +15,11 @@
       "description": "Whether to log in to workbench CLI",
       "proposals": ["true", "false"],
       "default": "false"
+    },
+    "shmSize": {
+      "type": "string",
+      "description": "Shared memory size for the application-server container",
+      "default": "64m"
     }
   }
 }
--- a/src/aou-sas/docker-compose.yaml
+++ b/src/aou-sas/docker-compose.yaml
@@ -4,6 +4,7 @@ include:
 services:
   app:
     container_name: "application-server"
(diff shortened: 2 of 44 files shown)


june-hua (approved):
Thanks! Is the plan to dynamically set it depending on machine type or have the user specify?

pantherman594:
> Thanks! Is the plan to dynamically set it depending on machine type or have the user specify?

For now I just wanted to do the minimum amount of work to make it user configurable, we can figure out how to set it dynamically/configurable in the UI in the future

cbirchsy:
Hi @pantherman594, 

Apologies if this is not the correct place to ask for help. 

I'm a user on the AoU verily workbench and I'm trying to do multi-GPU ML training but I'm severely bottlenecked by the 64mb shared memory. This PR looks like it could fix that issue, but when I use the above curl command and restart the app the shm size is unchanged. Would it be possible to confirm a step-by-step set of actions to change the shm-size on AoU JupyterLab VMs?

Thanks!

pantherman594:
Hey @cbirchsy, this change is still working its way up our environments, it should be in production later this week later this week (maybe Wednesday or Thursday)!

PR #450, opened 2026-07-20
PHP-169182: Transition to using gcloud storage

june-hua:
Starting March 2027, Google Cloud will no longer include the gsutil command-line tool in the default Google Cloud CLI installation package. They strongly encourage all users to transition to gcloud storage. See notice mentioned in [this gsutil doc](https://docs.cloud.google.com/storage/docs/gsutil_install).

All our scripts should update to use gcloud storage instead of gsutil. Verified that `gcloud storage` ls, cat, cp all exist, see [here](https://docs.cloud.google.com/sdk/gcloud/reference/storage).

[Explanation of differences ](https://docs.cloud.google.com/storage/docs/gsutil-transition-to-gcloud) between these two commands.

### DIFF
--- a/features/src/llm-context/generate-context.sh
+++ b/features/src/llm-context/generate-context.sh
@@ -375,7 +375,7 @@ wb resource add-ref aurora-database --name external-db
 
     else
         storage_bucket_type="GCS bucket"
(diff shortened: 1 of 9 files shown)


michaelxiao7 (approved):
lgtm

pantherman594 (approved):

PR #452, opened 2026-07-21
Application-server container memory limits

june-hua:
To prevent the application-server container from using all of the machine's memory, set a memory limit on the container.

### DIFF
--- a/startupscript/butane/050-parse-devcontainer.sh
+++ b/startupscript/butane/050-parse-devcontainer.sh
@@ -18,6 +18,105 @@ function usage {
   exit 1
 }
 
+calculate_container_memory_limit() {
+    # Query total system memory in bytes
+    local total_mem_bytes
+    total_mem_bytes=$(awk '/^MemTotal:/ {print $2 * 1024}' /proc/meminfo)
+    local total_mem_mb=$((total_mem_bytes / 1024 / 1024))
+
+    # Reserve memory for host OS, proxy-agent, fluent-bit, and other system processes
+    # Reserve whichever is larger: 1GB or 10% of total memory
+    local reserved_10pct=$((total_mem_mb * 10 / 100))
+    local reserved_mb=$((reserved_10pct > 1024 ? reserved_10pct : 1024))
+
+    # Container gets total - reserved
+    local container_mem_limit_mb=$((total_mem_mb - reserved_mb))
+
+    # Validate that the container memory limit is not negative
+    # If negative or zero, don't set a memory limit (let container use available memory)
+    if [[ "${container_mem_limit_mb}" -le 0 ]]; then
+        echo "WARNING: Calculated container memory limit is ${container_mem_limit_mb}MB (negative or zero)" >&2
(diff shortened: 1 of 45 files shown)


startupscript/butane/050-parse-devcontainer.sh
```diff
@@ -95,75 +200,20 @@ if [[ ! "${SHM_SIZE}" =~ ^[0-9]+[bBkKmMgG][bB]?$ ]]; then
...
-    # Removes the application-server container if any value has changed since last run.
-    local rebuild=false
+# Calculate memory limit for application-server container
+CONTAINER_MEM_LIMIT=$(calculate_container_memory_limit)
 
-    if [[ ! -f "${CONTAINER_STATE_FILE}" ]]; then
-        echo "First run, initializing container state"
-        rebuild=true
+# Allow override via metadata
+MEMORY_LIMIT_OVERRIDE="$(get_metadata_value "memory-limit" "")"
```
pantherman594: do we not want to let the user configure the memory limit? since I believe they can set guest attributes but not always metadata, and we're not triggering a recreate if it changes. that's fine if so, but just wanted to confirm

pantherman594 (approved):

startupscript/butane/050-parse-devcontainer.sh
june-hua: Good callout, our naming isn't accurate, `set_metadata` sets the guest attributes and `get_metadata` gets the metadata.
Yes we should be using guest attributes, I'm leaning towards removing the MEMORY_LIMIT_OVERRIDE as we do require allocating a certain percentage of memory for system processes and other containers. Thoughts?

pantherman594: Yeah I think this is something we can just set ourselves and not let users change

startupscript/butane/050-parse-devcontainer.sh
```diff
@@ -18,6 +18,111 @@ function usage {
...
 }
 
+calculate_container_memory_limit() {
+    # Query total system memory in bytes
+    local total_mem_bytes
+    total_mem_bytes=$(awk '/^MemTotal:/ {print $2 * 1024}' /proc/meminfo)
+    local total_mem_mb=$((total_mem_bytes / 1024 / 1024))
+
+    # Reserve memory for host OS, proxy-agent, fluent-bit, and other system processes
+    # Reserve whichever is larger: 1GB or 10% of total memory
```
pantherman594: I wonder if we could just always reserve 1GB, I don't think the host should be using more than that

jtsao2 (approved):
ML ops stamp for mem limit, LGTM. Deferring to WB squad on script logic.

startupscript/butane/050-parse-devcontainer.sh
june-hua: With all the other containers as well, I wanted some buffer room. I updated to cap at 4GB so we're not taking up too much memory for large machine types

PR #456, opened 2026-07-29
Add data collection access logging app

kvo3:
Adds a new Workbench devcontainer app for auditing data collection access grants. The app is a Flask web application that queries BigQuery monitoring tables (`workbench-bq-log-sink`) to provide visibility into who has access to a data collection, how they got it (directly or via group membership), and when.

  Views

  - Forensic: Data Collection (landing page) — Full history of access grants for a data collection. Shows GROUP, MEMBER, and INDIVIDUAL grants with role,
  timestamp, and actor. Group rows are expandable to show who was in the group at the time of the grant. Supports hiding revoked access and toggling between
  user-facing and internal group names.
  - Group Membership Audit — Full grant/revoke history for a Workbench group, with a "current members only" filter.

  Features

  - Configurable org and environment via env vars, config.yaml, or auto-detected from wb workspace describe
  - Org override field on all views for cross-org lookups without restarting the app
  - Search values persist across tab navigation
  - Client-side table filtering (AND/OR keyword search), column sorting, column-level filters, and resizable columns

  Infrastructure

  - Dockerized with docker-compose.yaml and .devcontainer.json for Workbench devcontainer deployment
  - Python 3.11 slim base image, Gunicorn with 2 workers / 4 threads
  - .gitignore added to exclude .venv/, __pycache__/, and .env files

BENCH-8927


emmarogge (approved):
LGTM!

yonghaoy:
Hey @kvo3 that is a custom app, customers are allowed to bring whatever app into workbench, as long as the app is not a built-in default 1st party app, no review is required.

However, for the purpose of the app, building a app just for DC audit is a bit overkill? It cost time and money to run that app, why not just query from BQ directly? 

For AoU, I just provide a snippet of python code and let them to run that code in Jupyter

NavidZ:
@yonghaoy you are correct. Basically at the end this app (or that script you provided to folks) is a bunch of sql queries that folks can run.
This app just gives it a UI and run those queries behind the scene and render it in the UI in a nice way. instead of you giving folks python code to run in a jupyter they would run this app directly (if they choose to) with ideally a nicer UI. of course they can continue using their own direct access on the BQ and this doesn't prevent them from doing that direct path if they want to. This app targets more like a no-code way of exploring those audit logs.

src/dc-access-logging-app/app/app.py
```diff
@@ -0,0 +1,1840 @@
...
+
+def _parse_wb_workspace():
+    result = {}
+    try:
+        out = subprocess.run(
+            ["wb", "workspace", "describe"],
+            capture_output=True, text=True, timeout=10,
+        )
+        for line in out.stdout.splitlines():
+            for key in ("Organization:", "Google project:"):
```
pantherman594, line 22: In the future it might be easier to use --format json and read the output as json

src/dc-access-logging-app/app/verily-workbench-style-guide.html
pantherman594, line 1: do we need to include this file?

pantherman594 (approved):
LGTM, left some small comments but none are blockers

src/dc-access-logging-app/app/app.py
```diff
@@ -0,0 +1,1840 @@
...
+from flask import Flask, jsonify, render_template_string, request
+from flask_cors import CORS
+from google.cloud import bigquery
+
+
+def _parse_wb_workspace():
+    result = {}
+    try:
+        out = subprocess.run(
+            ["wb", "workspace", "describe"],
```
june-hua, line 18: does this work? I don't see `wb` installed

src/dc-access-logging-app/app/app.py
```diff
@@ -0,0 +1,1840 @@
...
+        for row in rows:
+            for k, v in row.items():
+                if hasattr(v, 'isoformat'):
+                    row[k] = v.isoformat()
+        return jsonify(rows)
+    except Exception as e:
+        return jsonify({"error": str(e)}), 500
+
+
+@app.route("/health")
```
june-hua, line 1834: is this redundant?

june-hua (approved):

PR #464, opened 2026-08-21
Add virtual-browser templates

june-hua:
Based off changes in #463. Adds two app templates that serve a backend app (JupyterLab / RStudio)
through a server-side Chromium session rendered by [Selkies](https://github.com/selkies-project) and
streamed to the client as pixels, sharing one hardened browser front end.

## Templates

- **`virtual-browser-jupyter`** — JupyterLab through the streamed browser.
- **`virtual-browser-rstudio`** — RStudio Server through the streamed browser.

Each runs two containers on an internal `app-network`: `application-server` (the Chromium/Selkies
front end, the only published container, port `3000`) and the backend (`jupyterlab:8888` /
`rstudio:8787`, internal only). Chromium opens the backend in `--kiosk`.

## Shared browser front end (`src/browser-common/`)

Both templates share one image. Compose `include:` can't override an imported service, so each
template defines its own `app` service pointing `build.context` at `../browser-common` and passing an
`APP_ORIGIN` build arg.

- **`browser.Dockerfile`** — pinned Chromium base + Selkies/hardening env; generates the managed
  policy from the template at build.
- **`managed-policy.json.tmpl`** — managed Chromium enterprise policy; `__APP_ORIGIN__` is
  substituted with the template's origin.

### Managed Chromium policy

Baked into `/etc/chromium/policies/managed/workbench-rbi.json` (immutable; baked, not mounted, since
the workspace mount is user-writable). `URLAllowlist` = the single app origin, `URLBlocklist` =
`["*"]` (fails closed). Also disables downloads/export, file dialogs/file-system access,
clipboard-out, printing, devtools + remote debugging, extensions, new tabs / pop-ups / off-app
navigation, incognito/extra profiles, and password manager / autofill / translation / notifications.

### Selkies hardening (image env)

Sidebar hidden; sharing/collab disabled; clipboard-out and command disabled and `|locked`; file
transfers upload-only; audio/mic off; framerate/encoding tuned.

## Base image pinning

`lscr.io/linuxserver/chromium@sha256:4c7b9086…` with a `feature-versions/state.json` entry so
`update.sh` tracks it (dedicated `src/.*/browser\.Dockerfile` filter, since the existing
`Dockerfile` filters don't match).

## Tests & CI

Shared `tests/common/virtual-browser.bats` (parameterized by `BACKEND_CONTAINER` / `APP_ORIGIN`),
per-template runners, and both templates registered in `test-pr.yaml`.

## Embedding requirement (important)

The streamed client only starts its video pipeline on load if the **host page** embeds it with an
autoplay-capable permissions policy:

```html
<iframe … allow="autoplay; fullscreen; clipboard-read; clipboard-write; screen-wake-lock"></iframe>
```

Without this (cross-origin iframe), the app shows a black screen until the user switches tabs and
back. This is a host-page requirement, not a container setting — **no Selkies/pixelflux env changes
are needed** (earlier `H264_STREAMING_MODE` / `PIXELFLUX_WAYLAND` experiments were removed as
unnecessary once the iframe permissions were fixed). The host-side fix lives in verily1 (adds these
to the app iframe's `allow`). See the templates' README "Embedding" sections.


src/virtual-browser-jupyter/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+        # Baked into the policy URLAllowlist; must match the CHROME_CLI URL below.
+        APP_ORIGIN: "http://jupyterlab:8888"
+    restart: always
+    ports:
+      - "3000:3000"
+    networks:
+      - app-network
+    environment:
+      # --kiosk: fullscreen, no window chrome/decorations, no address bar.
+      CHROME_CLI: "--kiosk --proxy-server=0.0.0.0:0 --proxy-bypass-list=jupyterlab http://jupyterlab:8888"
```
june-hua: Still looking into how to fix fullscreen mode (--kiosk) , which causes a mouse-click offset on initialization. Currently, only resizing the window fixes this.

We could either not do full screen and find another way to hide the close and resize buttons on the toolbar, or resize in the workbench frontend after load.

src/virtual-browser-jupyter/.devcontainer.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "JupyterLab in a Virtual Browser devcontainer template",
+  "dockerComposeFile": ["docker-compose.yaml", "../jupyter-common/jupyter-common-compose.yaml"],
```
pantherman594, line 3: Do we need to add the browser docker compose file here? I don't remember exactly what it's for but I think there's a part of the devcontainer build that might need it

src/virtual-browser-jupyter/docker-compose.yaml
pantherman594: Allowing the user to close the window may not be the worst thing, since RESTART_APP is enabled it should quickly reopen the window. It could be helpful if their page gets in a bad state and they want to reload it without restarting the app.

Additionally the virtual browser can't be opened multiple times, so if someone wanted to open their app in multiple tabs they could do it in the virtual browser instead

pantherman594: do we still need the proxy-server/proxy-bypass if we're using the policy URLBlockList?

src/virtual-browser-jupyter/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+  # Browser front end. Built from the shared ../browser-common image; this template supplies only
+  # the JupyterLab-specific values. Selkies + hardening defaults are baked into that image.
+  app:
+    container_name: "application-server"
+    build:
+      context: ../browser-common
+      dockerfile: browser.Dockerfile
+      args:
+        # Baked into the policy URLAllowlist; must match the CHROME_CLI URL below.
+        APP_ORIGIN: "http://jupyterlab:8888"
```
pantherman594: we might need to omit the scheme, otherwise websockets might not work
```suggestion
        APP_ORIGIN: "jupyterlab:8888"
```

src/browser-common/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    NO_GAMEPAD=true \
+    HARDEN_DESKTOP=true \
+    HARDEN_KEYBINDS=true \
+    DISABLE_CLOSE_BUTTON=true \
+    RESTART_APP=true \
+    SELKIES_FRAMERATE=20 \
+    SELKIES_H264_CRF=28 \
+    SELKIES_AUDIO_ENABLED=false \
+    SELKIES_MICROPHONE_ENABLED=false \
+    SELKIES_UI_SHOW_SIDEBAR="false|locked" \
```
pantherman594: If the sidebar is completely hidden, users won't be able to upload files

@yonghaoy what level of isolation do we want? No in/out at all, or only block out (uploading is fine, but downloading is blocked). Potential downside is it could be easier to upload in an exploit if one is found

src/browser-common/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Selkies + hardening defaults; a template may override any of these in its compose environment.
+ENV SELKIES_DESKTOP=false \
+    NO_GAMEPAD=true \
+    HARDEN_DESKTOP=true \
+    HARDEN_KEYBINDS=true \
+    DISABLE_CLOSE_BUTTON=true \
+    RESTART_APP=true \
+    SELKIES_FRAMERATE=20 \
+    SELKIES_H264_CRF=28 \
+    SELKIES_AUDIO_ENABLED=false \
```
pantherman594, line 15: Do we want to support screen reader extensions for accessibility? I see that all extensions are blocked too, so maybe not

src/virtual-browser-jupyter/docker-compose.yaml
```diff
@@ -0,0 +1,46 @@
...
+      additional_contexts:
+        jupyter-extension-builder: service:jupyter-common-extension-builder
+    user: "jupyter"
+    shm_size: "${templateOption:shmSize}"
+    mem_limit: "${templateOption:memoryLimit}"
+    restart: always
+    volumes:
+      - .:/workspace:cached
+    networks:
+      - app-network
```
pantherman594: I wonder if we should actually put this on its own separate network, since otherwise app proxy agent could technically reach it

Proxy agent: network A
virtual browser: networks A, B
jupyterlab: network B

src/browser-common/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+ARG APP_ORIGIN=""
+
+# Chromium applies every *.json under /etc/chromium/policies/managed/ at startup. The policy is baked
+# into the image, not bind-mounted: the app container's workspace mount is user-writable, so a mounted
+# policy could be edited from inside the session.
+COPY managed-policy.json.tmpl /tmp/managed-policy.json.tmpl
+RUN mkdir -p /etc/chromium/policies/managed \
+    && sed "s|__APP_ORIGIN__|${APP_ORIGIN}|g" /tmp/managed-policy.json.tmpl \
+       > /etc/chromium/policies/managed/workbench-rbi.json \
```
pantherman594, line 49: could we write this file as root and make it read only (chmod 444)? Otherwise the user could potentially write to it

yonghaoy (approved):
can't wait to see the demo!

src/virtual-browser-rstudio/.devcontainer.json
```diff
@@ -0,0 +1,48 @@
...
+  "postStartCommand": [
+    "./startupscript/remount-on-restart.sh",
+    "rstudio",
+    "/home/rstudio",
+    "${templateOption:cloud}",
+    "${templateOption:login}"
+  ],
+  "features": {
+    "ghcr.io/devcontainers/features/java@sha256:8157bab2d8d71e40b2f3128c162fab763e2b11038fdd33784549290a5d386b48": {
+      "version": "17"
```
pantherman594: we should update this to use the temurin version as well, like in https://github.com/verily-src/workbench-app-devcontainers/pull/472

tests/common/virtual-browser.bats
```diff
@@ -0,0 +1,66 @@
...
+    run docker exec "${BROWSER_CONTAINER}" printenv SELKIES_COMMAND_ENABLED
+    [ "${status}" -eq 0 ]
+    [ "${output}" = "false|locked" ]
+}
+
+@test "file transfers are upload-only" {
+    run docker exec "${BROWSER_CONTAINER}" printenv SELKIES_FILE_TRANSFERS
+    [ "${status}" -eq 0 ]
+    [ "${output}" = "upload" ]
+}
```
pantherman594, line 72: not sure how valuable these test cases are, they're just checking static values that we set in the dockerfile

src/browser-common/Dockerfile
pantherman594, line 1: nit: since this is the only Dockerfile in this directory, it could probably be just `Dockerfile`

src/virtual-browser-jupyter/.devcontainer.json
june-hua: I didn't have any issues with this. The browser is configured in the docker-compose file in the same dir.

src/virtual-browser-jupyter/docker-compose.yaml
june-hua: Would that be an issue? The user shouldn't be able to change where the agent forwards the traffic. Did you have other concerns?

src/browser-common/Dockerfile
june-hua: deferring this to a future PR

src/virtual-browser-jupyter/docker-compose.yaml
june-hua: yep can be removed

src/virtual-browser-rstudio/.devcontainer.json
june-hua: thanks, just rebased to bring in your fix

src/virtual-browser-jupyter/docker-compose.yaml
pantherman594: I guess it's more of a defense-in-depth, especially when we have shareable apps where the port is configurable. But we're not exposing the port either so maybe not an issue if the user can't change the host?

src/virtual-browser-rstudio/.devcontainer.json
pantherman594: oh I think you might need to manually add it in, since this file didn't exist upstream

src/virtual-browser-jupyter/docker-compose.yaml
pantherman594: I tried visiting a blocked page (`echo https://google.com` in the terminal, it renders a clickable link) and I get this with no way to get out, even closing and reopening gets me back to the same place

image

tests/common/virtual-browser.bats
```diff
@@ -0,0 +1,66 @@
...
+# container (application-server) and that the backend app container is up. Reads the policy on the
+# host with jq via `docker exec ... cat`, so the browser image needs no jq of its own.
+
+setup_file() {
+    echo "# Running ${BATS_TEST_FILENAME##*/}" >&3
+}
+
+BROWSER_CONTAINER="application-server"
+# Backend container name + app origin differ per template; the per-template .sh exports them.
+BACKEND_CONTAINER="${BACKEND_CONTAINER:-jupyterlab}"
```
pantherman594: I don't think we should have defaults for these, might make more sense to have the caller always set them

src/virtual-browser-jupyter/docker-compose.yaml
june-hua: That's a good find, we could write an extension to block this. But given the existing issue with click offset for cold starts, let's skip fullscreen/kiosk mode for now as you suggested. As you mentioned, the ux is fine and chromium has managed policies.

The kiosk mode offset bug seems to be in Selkies/Pixelflux side that may be fixed upstream, and we can revisit this when chromium  updates their selkies version.

src/virtual-browser-jupyter/docker-compose.yaml
june-hua: sg added as a precaution

tests/common/virtual-browser.bats
june-hua: update them!

src/virtual-browser-rstudio/.devcontainer.json
june-hua: whoops, thanks!

src/browser-common/Dockerfile
june-hua: spoke offline, we should allow upload, configured selkies toolbar to still be visible to allow upload

june-hua: fyi use named volume to expose uploads that are stored in chromium container to the jupyterlab (app) container.
This would help us avoid giving user permissions to browse the entire chromium file system.

pantherman594 (approved):
Looks great, thanks!

src/virtual-browser-jupyter/docker-compose.yaml
june-hua: In --app mode, we can't remove the minimize button. If we minimize the app, the app can not be opened again.
Thus, we decided to go with --kiosk mode. The click offset issue is handled with iframe resize. The issue that involves opening a blocked page with no close button, this can be fixed with Ctrl+W a chromium keybind.

PR #469, opened 2026-08-24
Report cache build status before stopping docker

michaelxiao7:
The cache build reports its final status by calling set_metadata, which on GCP is a curl to the metadata server but on AWS tags the EC2 instance by running the AWS CLI in a container, so it cannot run after this script has masked and stopped the docker daemon.

Move the status write ahead of the docker stop so the AWS cache builder can tell a finished build from one that died, since it reads that status once the instance has powered itself off.

This changes nothing on GCP, where the call never depended on docker.

PHP-151845


pantherman594 (approved):
Hmm we're in a weird situation here. We want to clean up unnecessary files before saving the cache image, which we need to stop docker to do. But we need docker to write to the metadata. I think it should be fine since the signal to the cache builder is the instance shutting down, not the metadata getting set

PR #473, opened 2026-08-24
Add OpenCode + NVIDIA Nemotron custom app

samhornstein:
Adds `src/opencode-nemotron`: code-server with the [opencode](https://opencode.ai) agent driving a local NVIDIA Nemotron model on Ollama. Nothing leaves the VM. For the NVIDIA dev day demo. Derived from `feature/applegath/ollama_gemma-4-26B-A4B`.

This PR:
* Serves code-server on 8443 and Ollama on 11434 (container-local). Pins `opencode` 1.18.22.
* Defaults to `nemotron-3.5-lightning:30b` — 30B MoE, tool calling, 25 GB of weights. Needs an A100 40 GB.
* Resolves the model tag in `resolve-model.sh`: `/config/.opencode-model` first, then `OLLAMA_MODEL` from `docker-compose.yaml`. Both `start-ollama.sh` and `configure-opencode.sh` call it, so the server and the agent config cannot drift.
* Writes `~/.config/opencode/opencode.json` on create and restart, with `share: "disabled"` so sessions never reach opencode's hosted sharing service.
* Sets `OLLAMA_CONTEXT_LENGTH=32768`. Tool calls are unreliable below ~16k.

Two deviations worth review:

* **The model tag is not a template option.** The VM substitutes only `login`, `cloud`, `containerImage`, `containerPort`, `shmSize`, and `memoryLimit` (`startupscript/butane/050-parse-devcontainer.sh`). A custom option stays literal and breaks `docker-compose config`. So the default lives in `OLLAMA_MODEL` in `docker-compose.yaml`. To change the model on an existing app, write the tag to `/config/.opencode-model` and restart. `/config` is a volume, so the override survives a restart and a machine-type change. An A100 to L4 move needs a smaller model, and this is how you make that change.
* **OpenJDK 17 comes from apt, not `ghcr.io/devcontainers/features/java`.** That feature is currently broken for all 13 apps that use it, including `vscode-with-llm` and `nemo_jupyter`: Microsoft's Java 17 SDKMAN identifier is now `17.0.20+1-ms`, and the feature's regex rejects the `+1` segment. Upstream: devcontainers/features#1712, fix in #1713. Pinning the digest does not help — the lookup is against SDKMAN's live catalog. The other apps need a separate ticket. This app stays unaffected by the upstream outcome either way; `install-java.sh` only needs `/usr/bin/java`.

## Testing

Deployed to `samh-sandbox` on `a2-highgpu-1g` (1x A100 40 GB, us-central1-c):

* `ollama ps` → `nemotron-3.5-lightning:30b`, 25 GB, **100% GPU**, context 32768.
* `opencode run "Read hello.txt and tell me exactly how many lines it has"` → agent called its read tool and answered correctly.
* `java` 17.0.19 and `wb` on the `PATH`, so `post-startup.sh` succeeds. code-server answers on 8443.
* Model override → `resolve-model.sh` returns `nemotron-3.5-lightning:30b` by default and `nemotron-3-nano:4b` with `/config/.opencode-model` present. `configure-opencode.sh` then writes `ollama/nemotron-3-nano:4b`. Deleting the file reverts to the default. The `abc` user can write `/config` directly, so the documented command needs no `sudo`.
* Direct API example → with `max_tokens=100` the reply stops at `finish_reason=length` and `content` is empty. With no cap, `finish_reason=stop` and `content` is `'Hello! How can I help you today?'`. Nemotron always returns its thinking in Ollama's non-standard `reasoning` field, so the README prints both.

Not tested: `ollama pull` of a changed tag on restart. That path is a single line and reads the verified resolver.

Not registered in `test-pr.yaml`: CI runs `devcontainer up`, which would pull 25 GB on a GPU-less runner.

BENCH-9850

### DIFF
--- a/src/opencode-nemotron/README.md
+++ b/src/opencode-nemotron/README.md
@@ -51,12 +51,17 @@ print(response.choices[0].message.content)
 
 ## Changing the Model
 
-Set the `model` template option when you create the app. Use any tag from
-https://ollama.com/library. The value must be a tag that Ollama can pull, and the
-model must support tools for `opencode` to work.
+The model tag lives in `OLLAMA_MODEL` in `docker-compose.yaml`. Edit it in your
+fork and point the app config at that branch. Workbench substitutes only a fixed
+set of template options on the VM, so a custom `model` template option would not
+work.
 
-To change the model on a running app, edit `~/.config/opencode/opencode.json` and
-run `ollama pull <tag>`. The app rewrites that file on each restart.
+Use any tag from https://ollama.com/library. The model must support tools, or
+`opencode` cannot call them.
+
+To change the model on a running app, run `ollama pull <tag>` and edit
+`~/.config/opencode/opencode.json`. The app rewrites that file on each restart, so
+the change lasts until the next restart.
 
 ## Configuration
 
--- a/src/opencode-nemotron/README.md
+++ b/src/opencode-nemotron/README.md
@@ -56,17 +56,28 @@ print(message.content)
 
 ## Changing the Model
 
-The model tag lives in `OLLAMA_MODEL` in `docker-compose.yaml`. Edit it in your
-fork and point the app config at that branch. Workbench substitutes only a fixed
-set of template options on the VM, so a custom `model` template option would not
-work.
+Write the tag to `/config/.opencode-model`, then restart the app:
+
+```sh
+echo nemotron-3-nano:4b > /config/.opencode-model
+```
+
+`/config` is a volume, so the override survives a restart and a machine-type
+change. On restart the app pulls the new tag and rewrites
+`~/.config/opencode/opencode.json` to match. Delete the file to return to the
+default.
+
+Pick a model that fits the GPU. See the table above. A model larger than VRAM
+runs partly on the CPU and is very slow. Change the tag before you move an app to
+a smaller GPU, for example from an A100 to an L4.
 
 Use any tag from https://ollama.com/library. The model must support tools, or
 `opencode` cannot call them.
 
-To change the model on a running app, run `ollama pull <tag>` and edit
-`~/.config/opencode/opencode.json`. The app rewrites that file on each restart, so
-the change lasts until the next restart.
+To change the default for new apps, edit `OLLAMA_MODEL` in `docker-compose.yaml`
+in your fork and point the app config at that branch. Workbench substitutes only
+a fixed set of template options on the VM, so a custom `model` template option
+would not work.
 
 ## Configuration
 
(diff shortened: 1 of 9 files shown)


pantherman594:
If this is for a NVIDIA demo, would it be fine staying in/running from a branch? Does it need to be merged into master?

samhornstein:
> If this is for a NVIDIA demo, would it be fine staying in/running from a branch? Does it need to be merged into master?

@pantherman594 Technically speaking, we can pull in a custom app from a feature branch, so we don't *need* to merge into `master`. The question then is if we *want* this in `master` or not. Thoughts?

samhornstein:
**Decision:** Wait on merge until after demo

src/opencode-nemotron/README.md
```diff
@@ -0,0 +1,105 @@
...
+from openai import OpenAI
+
+client = OpenAI(base_url="http://localhost:11434/v1", api_key="unused")
+
+response = client.chat.completions.create(
+    model="nemotron-3.5-lightning:30b",
+    messages=[{"role": "user", "content": "Hello!"}],
+    max_tokens=100,
+)
+print(response.choices[0].message.content)
```
PeterSu92: did you test this? When I tried, using `response.choices[0].message.content` didn't work, but `response.choices[0].message.reasoning}` did - apparently this is a characteristic of the Nemotron model's output format?

src/opencode-nemotron/README.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+## Changing the Model
+
+The model tag lives in `OLLAMA_MODEL` in `docker-compose.yaml`. Edit it in your
```
PeterSu92: nit: it would be great if this was more configurable than having to edit the `docker-compose.yaml` file before app creation, especially bc you can change the config of a Workbnech app after it's stopped, so if you wanted to move from A100 -> L4, it'd break

src/opencode-nemotron/README.md
samhornstein: Good catch, and no — that snippet was not tested. The `max_tokens=100` cap truncates during the thinking phase, so generation never reaches the final answer and `content` stays empty while Ollama returns everything in the non-standard `reasoning` field.

Fixed in ee9cac6: dropped the cap and printed both fields.

I have not re-run it on the GPU VM yet. If `content` is still empty with no cap, then it is not truncation and I will document `reasoning` as the answer field instead.

src/opencode-nemotron/README.md
samhornstein: Agreed, and it is worse than a nit: after an A100 -> L4 move the app would still pull the 25 GB model onto a 24 GB card and silently fall back to CPU.

A template option cannot fix it, since the VM substitutes only six options. So 3a31f42 puts the override in the persistent volume instead:

```sh
echo nemotron-3-nano:4b > /config/.opencode-model
```

Restart the app. `/config` is a volume, so the override survives a stop, a start, and a machine-type change. Delete the file to return to the default.

New `resolve-model.sh` is the single source for the tag. Both `start-ollama.sh` and `configure-opencode.sh` call it, so the server and the agent config cannot drift.

src/opencode-nemotron/README.md
samhornstein: Confirmed on the GPU VM. It was truncation:

```
max_tokens=100   finish_reason=length   content=''
max_tokens=None  finish_reason=stop     content='Hello! How can I help you today?'
```

`reasoning` was populated in both cases. So `content` does hold the answer once generation completes — the old 100-token cap just stopped the model partway through its thinking. The example in ee9cac6 is correct as written.

src/opencode-nemotron/README.md
samhornstein: Verified on the GPU VM:

```
default       : nemotron-3.5-lightning:30b
with override : nemotron-3-nano:4b
opencode model: ollama/nemotron-3-nano:4b
after revert  : ollama/nemotron-3.5-lightning:30b
```

The override propagates to the opencode config, and deleting the file returns to the default. `abc` can write `/config` directly, so no `sudo` is needed for the documented command.

PR #477, opened 2026-09-01
Upgrade pgweb to a non-EOL debian image

june-hua:
Wiz scan failing because debian 11 is EOL. The pgweb docker image is still on debian 11. Build the image using debian 12 and extract pgweb into it.

### DIFF
--- a/feature-versions/state.json
+++ b/feature-versions/state.json
@@ -84,6 +84,11 @@
     "installed": "sha256:a5256d416e2e8b92d69a4459058e3eca33a9f075d8325491644411d0bc3bd70b",
     "filter": "src\\/.*\\/Dockerfile"
   },
+  "debian": {
+    "tag": "12.15-slim",
+    "installed": "sha256:88200866dfff7ea7f5cbcb6ec7c8a701889efe6fe859fe64d6990e4b07ea4171",
+    "filter": "src\\/pgweb\\/Dockerfile"
+  },
   "mikefarah/yq": {
     "tag": "4",
     "installed": "sha256:cfc4eee658595834ef304eadb0c3ea721f3b7cb6404ad8b7cb909cc5b5145b23",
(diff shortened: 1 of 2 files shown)


feature-versions/state.json
```diff
@@ -84,6 +84,11 @@
     "installed": "sha256:a5256d416e2e8b92d69a4459058e3eca33a9f075d8325491644411d0bc3bd70b",
     "filter": "src\\/.*\\/Dockerfile"
   },
+  "debian": {
+    "tag": "12.15-slim",
```
pantherman594: should we use debian 13-slim like the [pgweb image](https://github.com/sosedoff/pgweb/blob/e4858a16d8e032730055289596ea9059a91bca64/Dockerfile#L33)? (not sure why they haven't had a release in a while...)

src/pgweb/Dockerfile
```diff
@@ -1,12 +1,31 @@
...
 USER root
 
 RUN apt-get update && \
-    apt-get -y install sudo && \
+    apt-get -y install --no-install-recommends \
+        ca-certificates \
+        curl \
+        netcat-openbsd \
+        openssl \
+        postgresql-client \
```
pantherman594, line 31: pgweb image [uses](https://github.com/sosedoff/pgweb/blob/e4858a16d8e032730055289596ea9059a91bca64/Dockerfile#L38) the postgresql apt repo which has postgresql-client-18, debian 12 is [stuck on](https://packages.debian.org/search?keywords=postgresql-client#:~:text=PostgreSQL%20(supported%20version)-,15%2B248%2Bdeb12u1%3A%20all,-also%20provided%20by) 15 (and debian 13 is on postgres 17). should we do the same?

june-hua: good catch, updating!

pantherman594 (approved):
Thanks!

PR #485, opened 2026-09-16
bootstrap-files.sh skip download if file already exists

pantherman594:
So it can be rerun without redownloading


yonghaoy (approved):

june-hua (approved):
TY!

michaelxiao7 (approved):

